PR #37, opened 2021-11-06
Merge in remove identification mod.

ChrisPJohn:


ThyWoof (changes requested):
. No need to call Load() on viewers as this is a restart type one.
. I’m a big fan of { } in any block, even 1 liners. OpenSSL is a good example of an exploit caused by that.

ChrisPJohn:
> . No need to call Load() on viewers as this is a restart type one. . I’m a big fan of { } in any block, even 1 liners. OpenSSL is a good example of an exploit caused by that.

I generally prefer using braces as well. I just missed it this time.

Since it is possible to enable it during play with that quick call I think we should. It will prevent users from thinking the toggle doesn't work or getting frustrated needing to save and reload the game. It's a nice quality of life improvement. It's probably rare that folks who want to play with it on turn it off, and if they do, the text explains a restart is needed in that case.

PhilPJL:
```internal bool NoIdentification;```
Need to make it public or it won't serialize.

PhilPJL:
You don't need ```GetAllElements```
You can use ```foreach (var item in DatabaseRepository.GetDatabase<ItemDefinition>())```
and save an array copy.

ChrisPJohn:
> `internal bool NoIdentification;` Need to make it public or it won't serialize.

done

ChrisPJohn:
> You don't need `GetAllElements` You can use `foreach (var item in DatabaseRepository.GetDatabase<ItemDefinition>())` and save an array copy.

done

ThyWoof (approved):

PR #38, opened 2021-11-06
No attunement

PhilPJL:
Similar to 'no identify'.  

Apologies for tidying the code in the same commit.  Makes it hard to see what's changed.


ThyWoof (approved):

SolastaCommunityExpansion/Patches/GameUi/PowerSelectionPanelPatcher.cs
```diff
@@ -27,7 +27,7 @@ internal static void Postfix(PowerSelectionPanel __instance)
                 {
                     if (thirdRow == null)
                     {
-                        thirdRow = GameObject.Instantiate(powersTable);
+                        thirdRow = Object.Instantiate(powersTable);
```
ChrisPJohn, line 30: Is switching from GameObject to Object safe here (and elsewhere)?

SolastaCommunityExpansion/Models/RemoveAttunementRequirementContext.cs
```diff
@@ -0,0 +1,18 @@
+﻿using SolastaModApi.Extensions;
+
+namespace SolastaCommunityExpansion.Models
+{
+    internal static class RemoveAttunementRequirementContext
```
ChrisPJohn, line 5: Should we have a context just for this (and some other one line changes) or merge them together? I could see this combined with the no identification change.

ChrisPJohn (changes requested):
One question on the refactoring and one about how we want to organize going forward.

SolastaCommunityExpansion/Patches/GameUi/PowerSelectionPanelPatcher.cs
PhilPJL: Yes, I tend to trust VS refactorings.  Instantiate is implemented on Object not GameObject.

SolastaCommunityExpansion/Models/RemoveAttunementRequirementContext.cs
PhilPJL: Merged as suggested

ChrisPJohn (approved):
One question on the refactoring and one about how we want to organize going forward.

PR #40, opened 2021-11-06
Set all faction relations to max

PhilPJL:
Zappa to suitable UI

### DIFF
--- a/SolastaCommunityExpansion/Models/SetFactionRelationsContext.cs
+++ b/SolastaCommunityExpansion/Models/SetFactionRelationsContext.cs
@@ -1,0 +1,23 @@
+﻿namespace SolastaCommunityExpansion.Models
+{
+    internal static class SetFactionRelationsContext
+    {
+        internal static void Load()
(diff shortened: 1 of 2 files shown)


ThyWoof:
I might change it a bit if it makes any sense to add a setting per Faction with a corresponding toggle / slider on a viewer.

@ChrisPJohn , Should we have a new viewer "Merchants & Factions"? Should we integrate this with "Items & Recipes"?

ChrisPJohn:
I like having a slider per faction and a button to set all to max in order provide both flexibility for any that want it as well as the ease of one click max.

As for where in the UI, I think potentially a new cheats/trainer viewer might be the best place. There's some other things we can do with those kind of modifications (add gold, grant items, die roll modifications).

ThyWoof:
OK. Accept this one for now and I'll work on an UI once it gets to main

ThyWoof (dismissed):

ChrisPJohn (approved):

PR #41, opened 2021-11-06
migrates RESPEC from UB to CE

ThyWoof:
. Migrates RESPEC feature from Unfinished Business into Community Expansion

. Let's keep in sync as I need to update UB first on Nexus (without RESPEC) and CE after to avoid any issues


SolastaCommunityExpansion/Features/RestActivityRespecBuilder.cs
```diff
@@ -0,0 +1,27 @@
+﻿using SolastaModApi;
+using SolastaModApi.Extensions;
+using static SolastaModApi.DatabaseHelper.RestActivityDefinitions;
+
+namespace SolastaCommunityExpansion.Features
+{
+    public class RestActivityRespecBuilder : BaseDefinitionBuilder<RestActivityDefinition>
```
ChrisPJohn, line 7: Because this builder is specific to respec, I was wondering if a better spot would be in the respec context file instead of a shared folder.

SolastaCommunityExpansion/Patches/Respec/RulesetCharacterHeroPatcher.cs
```diff
@@ -0,0 +1,29 @@
...
+        [HarmonyPatch(typeof(RulesetCharacterHero), "EnumerateAfterRestActions")]
+        internal static class RulesetCharacterHero_EnumerateAfterRestActions
+        {
+            internal static void Postfix(List<RestActivityDefinition> ___afterRestActions)
+            {
+                if (Main.Settings.EnableRespec)
+                {
+                    foreach (var restActivityDefinition in DatabaseRepository.GetDatabase<RestActivityDefinition>().GetAllElements())
+                    {
+                        switch (restActivityDefinition.Condition)
```
ChrisPJohn, line 18: Is there a reason a switch with one case is better than an if here?

SolastaCommunityExpansion/Functors/FunctorRespec.cs
```diff
@@ -0,0 +1,196 @@
...
+        private const int RESPEC_STATE_ABORTED = 2;
+
+        private static int respecState = RESPEC_STATE_NORESPEC;
+        private static RulesetCharacterHero newHero;
+
+        public static bool IsRespecing => respecState == RESPEC_STATE_RESPECING;
+
+        private static readonly List<RulesetItemSpellbook> rulesetItemSpellbooks = new List<RulesetItemSpellbook>();
+
+        internal static void DropSpellbooksIfRequired(RulesetCharacterHero rulesetCharacterHero)
```
ChrisPJohn, line 22: Is there any harm in letting them keep the spell book?

If somebody had wizard levels and the respecs to another build that also contains wizard levels they may be frustrated that they lost all of the spells scribed into their spellbook.

SolastaCommunityExpansion/Functors/FunctorRespec.cs
```diff
@@ -0,0 +1,196 @@
...
+
+            gameCampaignScreen.Hide(true);
+            gameLocationScreenExploration?.Hide(true);
+            guiConsoleScreen.Hide(true);
+
+            characterBuildingService.CreateNewCharacter();
+            newHero = characterBuildingService.HeroCharacter;
+
+            functorParameters.RestingHero?.CharacterInventory?.BrowseAllCarriedItems<RulesetItemSpellbook>(rulesetItemSpellbooks);
+            DropSpellbooksIfRequired(functorParameters.RestingHero);
```
ChrisPJohn, line 74: If we are going to drop the spellbook (see above comment), shouldn't we only do it if the respec is being completed?

SolastaCommunityExpansion/Functors/FunctorRespec.cs
```diff
@@ -0,0 +1,196 @@
...
+                    rulesetCharacterHero.LoseItem(rulesetItemSpellbook, false);
+                }
+            }
+        }
+
+        internal static void PickupSpellbooksIfRequired(RulesetCharacterHero rulesetCharacterHero)
+        {
+            if (rulesetCharacterHero.ClassesHistory[rulesetCharacterHero.ClassesHistory.Count - 1].Name == "Wizard")
+            {
+                foreach (var rulesetItemSpellbook in rulesetItemSpellbooks)
```
ChrisPJohn, line 37: Do we actually want to grant multiple spellbooks to a character?

ChrisPJohn (changes requested):
Mostly questions around spellbooks and code organization.

SolastaCommunityExpansion/Patches/Respec/RulesetCharacterHeroPatcher.cs
ThyWoof: lol. I had level down there at some point.

ChrisPJohn (approved):

ChrisPJohn (approved):

PR #42, opened 2021-11-06
introduces a two-step patching process

ThyWoof:
. 2 new flags to control status:
- Main.Loaded = true when game finishes GameManager.BindPostDatabase
- Main.Enabled = true when all other patches were applied and models were loaded
- IsSafeToLoad() is true at the moment but idea is be able to use this to determine if safe to patch the game or not

### DIFF
--- a/SolastaCommunityExpansion/Main.cs
+++ b/SolastaCommunityExpansion/Main.cs
@@ -4,11 +4,13 @@ using System.IO;
 using System.Reflection;
 using UnityModManagerNet;
 using ModKit;
+using HarmonyLib;
 
 namespace SolastaCommunityExpansion
 {
     public class Main
     {
+        public static bool Loaded = false;
         public static bool Enabled = false;
         public static readonly string MOD_FOLDER = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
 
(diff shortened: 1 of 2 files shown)


ThyWoof:
I only left a comment on Harmony decoration at GameManagerPatcher. Willing to reduce conflict during merge. Please move anything that got merged on this file Postfix() and move over to Main.FinishLoading()

ThyWoof:
and of course delete GameManagerPatcher.cs as it won't be required anymore

ThyWoof:
image

Added a different mod ID on purpose with a PreLoad postfix.

SolastaCommunityExpansion/Main.cs
```diff
@@ -27,15 +30,17 @@ internal static bool Load(UnityModManager.ModEntry modEntry)
...
                 Logger = modEntry.Logger;
                 Mod = new ModManager<Core, Settings>();
-                Mod.Enable(modEntry, assembly);
-
                 Menu = new MenuManager();
-                Menu.Enable(modEntry, assembly);
 
-                Translations.Load(MOD_FOLDER);
+                Menu.Enable(ModEntry, assembly);
+                harmony.Patch(methodGameManagerBindPostDatabase, postfix: new HarmonyMethod(methodMainFinishLoading));
```
ChrisPJohn, line 43: What's the benefit of making this patch dynamic?

Is this just a proof of concept for how we can dynamically load patches for features?

ThyWoof: I’m just playing with it for now. We might need to have better control of the patches at some point and good to now we can control order, decide to enable / disable, etc.

ThyWoof:
closing this one as this approach can produce some crazy bugs in the end.

PR #64, opened 2021-11-09
enables cheat window during gameplay

ThyWoof:
@ChrisPJohn , not sure if best location for this toggle is under House Rules. We might need to consider a new tab for these...


ThyWoof:
image

ChrisPJohn (changes requested):
Move the setting (and the no experience to level up) to a new tab called "Tools"

ChrisPJohn (approved):

PR #66, opened 2021-11-10
Add a check for if we're doing narration and don't pause the game if so.

ChrisPJohn:


PhilPJL (approved):

PhilPJL:
Tested.  Zappas fix isn't needed if this one is applied.

PR #67, opened 2021-11-10
Add a new ui for setting faction relations.

ChrisPJohn:

### DIFF
--- a/SolastaCommunityExpansion/Viewers/Displays/ToolsDisplay.cs
+++ b/SolastaCommunityExpansion/Viewers/Displays/ToolsDisplay.cs
@@ -52,14 +52,10 @@ namespace SolastaCommunityExpansion.Viewers.Displays
 
             UI.Label("");
             UI.Label("Faction Relations");
-            UI.Label("");
-            
 
             bool flip = true;
             var service = ServiceRepository.GetService<IGameFactionService>();
             if (service != null) {
-                UI.Label("Does not grant achievements for faction relations, you can modify the faction relationship through gameplay after setting it here to gain the achievements.".red());
-                UI.Label("");
                 foreach (FactionDefinition faction in service.RegisteredFactions)
                 {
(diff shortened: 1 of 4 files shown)


ThyWoof:
. Instead of UI.AutoWidth() I suggest UI.Width(number) on sliders to move them closer to title strings.
. Consider making the red message italic and a bit smaller to avoid UI horizontal scrolling on default UI mod window sizes.

ChrisPJohn:
> . Instead of UI.AutoWidth() I suggest UI.Width(number) on sliders to move them closer to title strings. . Consider making the red message italic and a bit smaller to avoid UI horizontal scrolling on default UI mod window sizes.

I changed it so moving the sliders now triggers all the notifications and achievements (can get spammy with the faction change notifications if you drag them all over the place). So I removed the red message.

As for setting the width on the sliders it had no effect at all (UI.Width(number) had the same with as UI.AutoWidth(), no matter what number I specified).  I moved the names closer to the sliders by adding some padding.

SolastaCommunityExpansion/Viewers/Displays/ToolsDisplay.cs
```diff
@@ -49,6 +49,44 @@ internal static void DisplayTools()
...
+                    string title = Gui.Format(faction.GuiPresentation.Title);
+                    if (flip)
+                    {
+                        title = title.yellow();
+                    } else
+                    {
+                        title = title.white();
+                    }
+                    int intValue = service.FactionRelations[faction.Name];
+                    if (UI.Slider("                              " + title, ref intValue, faction.MinRelationCap, faction.MaxRelationCap, 0, "", UI.AutoWidth()))
```
ThyWoof, line 80: you could also use UI.Space(number) on cases like this but it would also require an UI.HorizontalScope context. keep in mind for more complex layouts

PR #68, opened 2021-11-11
Add support for new custom fighting styles including restrictions for…

ChrisPJohn:
… when they're valid.

### DIFF
--- a/SolastaCommunityExpansion/CustomFeatureDefinitions/CustomFightingStyle.cs
+++ b/SolastaCommunityExpansion/CustomFeatureDefinitions/CustomFightingStyle.cs
@@ -1,0 +1,85 @@
+﻿using SolastaModApi;
+using SolastaModApi.Diagnostics;
+using SolastaModApi.Extensions;
+using SolastaModApi.Infrastructure;
+using System.Collections.Generic;
+
+namespace SolastaCommunityExpansion.CustomFeatureDefinitions
+{
+    public interface ICustomFightingStyle
+    {
+        bool IsActive(RulesetCharacterHero character);
+    }
+
+    public delegate bool IsActiveFightingStyleDelegate(RulesetCharacterHero character);
+
+    public class CustomizableFightingStyle : FightingStyleDefinition, ICustomFightingStyle
+    {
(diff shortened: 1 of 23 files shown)


ThyWoof:
ok with changes but also asking @PhilPJL to take a look as this is when I hit a wall at Solasta modding

PhilPJL (commented):
What's meant by "The ModAPI doesn't handle adding extensions of things that aren't feature definitions well.".  Do you have an example of what's not working?  AddToDB may need to be regenerated/rebuilt.

PhilPJL (approved):
The code looks ok (although good practise says don't use public fields - general comment not just this change). 
Unfortunately I don't have sufficient knowledge to comment on adding fighting styles etc.

ChrisPJohn:
> What's meant by "The ModAPI doesn't handle adding extensions of things that aren't feature definitions well.". Do you have an example of what's not working? AddToDB may need to be regenerated/rebuilt.

So this builder is building type: CustomizableFightingStyle
When the BaseDefinitionBuilder in ModAPI tries to add that to the db it tries to get a db of type CustomizableFightingStyle which doesn't exist and then the ModAPI throws an error. I was able to previously add a custom class feature definition because FeatureDefinitions are checked for if they are a specific type of feature and then added to that specific FeatureType db, which is the support needed for custom class feature definitions.

I see two possible solutions here
1) add support for specifying the db to get added to which would in this case let me say yeah the class is CustomizableFightingStyle but really this is a FightingStyleDefinition add it to that db. If we do this we should still check that the provided object does extend the provided database class.
2) Instead of attempting to add to one db most of the time, check if the provided object is a valid type for every db the game has and add it to all of the appropriate dbs.

ThyWoof:
Is there any case in the game where 2) would be useful?

ChrisPJohn:
> Is there any case in the game where 2) would be useful?

Short answer is yes. There are several things that go into multiple databases currently. The most common of them is FeatureDefinitions go into the FeatureDefinition db and the specific feature definition type db.

ThyWoof (approved):

PR #70, opened 2021-11-12
Slower Progression

Czartchonn:
Any chance for a option where you can slow down the xp you get? Hate it when you reach the cap early in a game and it adds difficulty too. Maybe a slider or an option for 75% to 50% received.


ThyWoof:
Please use the issues tab for suggestions if you aren't pushing any code.

PR #95, opened 2021-11-14
cursor lock

ThyWoof:


PhilPJL (commented):
I like the fact that there's no impact on mouse click speed, unlike patching the cursor manager which will have a small effect.  However I don't like the fix having to be applied to all mods to work.

ThyWoof:
closing this one in favor of ImpPhil ModApi solution

PR #98, opened 2021-11-15
Prevent XP scaling when Caer_Cyflen_Quest_AfterTutorial is active

PhilPJL:


PhilPJL:
Tested L2 transition and L5 transition.  L2 allowed to proceed.  L5 scaled but doesn't block game.

ThyWoof (approved):

PR #121, opened 2021-11-20
Fix deprecated turn

ChrisPJohn:


SolastaCommunityExpansion/Feats/AcehighFeats.cs
```diff
@@ -325,9 +329,6 @@ protected RageFeatConditionBuilder(string name, string guid) : base(DatabaseHelp
                 Definition.Features.Add(RageDamageBonusAttackModifierBuilder.RageDamageBonusAttackModifier);
                 Definition.SetDurationType(RuleDefinitions.DurationType.Minute);
                 Definition.SetDurationParameter(1);
-
-
-                Definition.SetDurationType(RuleDefinitions.DurationType.Turn);
```
ThyWoof, line 330: double check this pls as it isn't following same fix pattern for others

ChrisPJohn: I just made a character with the feat and verified the rage properly lasts for 1 minute and then is removed.

SolastaCommunityExpansion/Subclasses/Fighter/Tactician.cs
```diff
@@ -62,7 +62,7 @@ protected static FeatureDefinitionPowerSharedPool Build(string name, string guid
             newEffectDescription.SetDifficultyClassComputation(RuleDefinitions.EffectDifficultyClassComputation.AbilityScoreAndProficiency);
             newEffectDescription.SavingThrowAbility = "Strength";
             newEffectDescription.HasSavingThrow = true;
-            newEffectDescription.DurationType = RuleDefinitions.DurationType.Turn;
+            newEffectDescription.DurationType = RuleDefinitions.DurationType.Round;
```
ThyWoof, line 65: assuming DurationParameter defaults to zero but wouldn't be better to make it clear here?

ThyWoof: What is your take @PhilPJL ?

PhilPJL: Your assumption is wrong, so if it needs to be zero better set it to zero
image

PhilPJL: Probably best to set all required properties explicitly.

PhilPJL: DurationParameter has a public setter so no need to use SetDurationParameter although no problem if you do since the extension wraps public properties when they exist

```
        public static T SetDurationParameter<T>(this T entity, int value)
            where T : EffectDescription
        {
            entity.DurationParameter = value;
            return entity;
        }
```

ThyWoof: @ChrisPJohn , should this be 1 or 0?

ChrisPJohn: I believe this duration is irrelevant. The prone condition lasts until the character stands up. I've tested all of these changes and they are all working as I believe they're intended to.

ThyWoof (approved):

PR #124, opened 2021-11-20
Add editorconfig

PhilPJL:
I've added a default '.editorconfig' file and the standard Microsoft .net analyzer package.
I've then gone through and fixed all remaining warnings and messages (except one on unused 'concentrationAffinity' which may indicate a bug).
Double click on the editorconfig file in the solution explorer to see the available options. We can enable more as required.

### DIFF
--- a/SolastaCommunityExpansion/.editorconfig
+++ b/SolastaCommunityExpansion/.editorconfig
@@ -1,0 +1,133 @@
+# To learn more about .editorconfig see https://aka.ms/editorconfigdocs
+###############################
+# Core EditorConfig Options   #
+###############################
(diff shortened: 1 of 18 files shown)


PhilPJL:
Note that
```
                Definition.AffinityGroups.Add(
                    new SavingThrowAffinityGroup
                    {
                        affinity = RuleDefinitions.CharacterSavingThrowAffinity.Advantage,
                        abilityScoreName = "Strength"
                    });
```
is now adding the SavingThrowAffinityGroup.  It wasn't previously.

ChrisPJohn (approved):

PR #126, opened 2021-11-20
fix reqRestart reference

ThyWoof:


ThyWoof:
didn't catch this one. dev won't compile without this...

ChrisPJohn (approved):

PR #128, opened 2021-11-21
updates ModKit

ThyWoof:


PhilPJL (commented):
If you add a nuget reference it should look like

```
	<ItemGroup>
	  <PackageReference Include="ModKit" Version="1.0.8" />
	</ItemGroup>

```

PhilPJL (approved):

PR #129, opened 2021-11-22
barbarian persistent rage / ranger feral senses

ThyWoof:


SolastaCommunityExpansion/Level20/Features/ConditionPersistentRageBuilder.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace SolastaCommunityExpansion.Level20.Features
+{
+    internal class ConditionPersistentRageBuilder : BaseDefinitionBuilder<ConditionDefinition>
+    {
+        private const string ConditionPersistentRagingName = "ZSConditionPersistentRaging";
+        private const string ConditionPersistentRagingGuid = "08e1a91049414d2cb3a247f8dae14b66";
+
+        protected ConditionPersistentRageBuilder(string name, string guid) : base(ConditionRaging, name, guid)
+        {
+            Definition.SetDurationType(RuleDefinitions.DurationType.UntilAnyRest);
```
ThyWoof, line 14: thoughts? nerfed it a bit...

SolastaCommunityExpansion/Level20/Features/PowerBarbarianPersistentRageBuilder.cs
```diff
@@ -0,0 +1,38 @@
...
+        private const string PowerBarbarianPersistentRageStartName = "ZSPowerBarbarianPersistentRageStart";
+        private const string PowerBarbarianPersistentRageStartGuid = "feea8426e38f49799628b3ceb0abe877";
+
+        protected PowerBarbarianPersistentRageBuilder(string name, string guid) : base(PowerBarbarianRageStart, name, guid)
+        {
+            Definition.SetGuiPresentation(new GuiPresentationBuilder("Power/&ZSPowerBarbarianPersistentRageStartDescription", "Power/&ZSPowerBarbarianPersistentRageStartTitle").Build());
+            Definition.SetOverriddenPower(DatabaseHelper.FeatureDefinitionPowers.PowerBarbarianRageStart);
+            Definition.EffectDescription.EffectForms[0].ConditionForm.SetConditionDefinition(ConditionPersistentRage);
+
+            //
```
ThyWoof, line 19: is this ok to be here?

SolastaCommunityExpansion/Level20/Features/ConditionPersistentRageBuilder.cs
ChrisPJohn: If you don't attack or take damage will the rage still end?

SolastaCommunityExpansion/Level20/Features/PowerBarbarianPersistentRageBuilder.cs
```diff
@@ -0,0 +1,38 @@
...
+using static SolastaModApi.DatabaseHelper.FeatureDefinitionPowers;
+
+namespace SolastaCommunityExpansion.Level20.Features
+{
+    internal class PowerBarbarianPersistentRageBuilder : BaseDefinitionBuilder<FeatureDefinitionPower>
+    {
+        private const string PowerBarbarianPersistentRageStartName = "ZSPowerBarbarianPersistentRageStart";
+        private const string PowerBarbarianPersistentRageStartGuid = "feea8426e38f49799628b3ceb0abe877";
+
+        protected PowerBarbarianPersistentRageBuilder(string name, string guid) : base(PowerBarbarianRageStart, name, guid)
```
ChrisPJohn, line 13: So my understanding of how rage works currently is rage start and rage stop are actions that trigger a power. I don't believe this is going to work without updating what the action does. But that gets complicated because you need to only change the activated power when a barbarian of the proper level activates rage.

SolastaCommunityExpansion/Level20/Features/PowerBarbarianPersistentRageBuilder.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            Definition.SetGuiPresentation(new GuiPresentationBuilder("Power/&ZSPowerBarbarianPersistentRageStartDescription", "Power/&ZSPowerBarbarianPersistentRageStartTitle").Build());
+            Definition.SetOverriddenPower(DatabaseHelper.FeatureDefinitionPowers.PowerBarbarianRageStart);
+            Definition.EffectDescription.EffectForms[0].ConditionForm.SetConditionDefinition(ConditionPersistentRage);
+
+            //
+            // NOTE: please review if ok to do this change here. need to ensure that the stop power also check for persistent rage condition
```
ChrisPJohn, line 20: Rage stop currently specifically removes ConditionRaging, so a barbarian using this will not be able to stop raging.

ChrisPJohn (changes requested):
Have you tested this? Looking at the code I don't believe it will work as intended.

SolastaCommunityExpansion/Level20/Features/ConditionPersistentRageBuilder.cs
ThyWoof: `  public void StartBattleTurn()
  {
    if (!this.Valid)
      return;
    if (!this.hasAttackedSinceLastTurn && !this.hasReceivedDamageSinceLastTurn)
    {
      RulesetCondition activeCondition = (RulesetCondition) null;
      if (this.RulesetCharacter.TryGetConditionOfCategoryAndType("11Effect", "ConditionRaging", out activeCondition))
        this.RulesetCharacter.RemoveCondition(activeCondition);
    }
    this.ResetVariables();
    this.RefreshEffectsForBattleTurnStart();
    if (!this.RulesetCharacter.IsDeadOrDyingOrUnconscious)
      this.RefreshActionPerformances();
    foreach (RulesetActor controlledEffectProxy in this.RulesetCharacter.ControlledEffectProxies)
      GameLocationCharacter.GetFromActor(controlledEffectProxy)?.StartBattleTurn();
  }`

it looks like this one might require patching

ChrisPJohn: Because you make a new condition and don't use ConditionRaging, I believe your raging condition would stay on even if they don't attack or take damage.

There's still the problem I mentioned before that rage stop will not function currently and that both rage start and rage stop are specifically activated as an action that is linked to an explicit power which I think we'd need to patch to be able to change.

ChrisPJohn:
It may be difficult now, but in the future I'd suggest doing each new feature in their own branch in case one of them runs into roadblocks like the barbarian rage seems to be.

SolastaCommunityExpansion/Level20/Features/ConditionPersistentRageBuilder.cs
ThyWoof: I’m researching alternatives to avoid patching

ThyWoof:
closing this one. will open 3 individual requests to avoid roadblocks

PR #131, opened 2021-11-22
character export first commit

ThyWoof:


ThyWoof:
image

ThyWoof:
image

ThyWoof:
image

SolastaCommunityExpansion/Patches/CharacterExport/MessageModalPatcher.cs
```diff
@@ -0,0 +1,35 @@
...
+using TMPro;
+using UnityEngine;
+
+namespace SolastaCommunityExpansion.Patches
+{
+    internal static class MessageModalPatcher
+    {
+        internal static TMP_InputField InputField { get; private set; }
+
+        [HarmonyPatch(typeof(MessageModal), "Show")]
```
ChrisPJohn, line 11: Where else is MessageModal used, is this going to add an input field to every modal? Should this be an extension method on MessageModal that you call only from where you need input?

SolastaCommunityExpansion/Models/CharacterExportContext.cs
```diff
@@ -0,0 +1,236 @@
...
+                }
+                catch (Exception ex)
+                {
+                    Main.Log($"RegisterEntity: {ex}");
+                }
+            }
+
+            public bool TryGetEntityByGuid(ulong guid, out RulesetEntity rulesetEntity)
+            {
+                Main.Log($"TryGetEntityByGuid: {guid}");
```
ChrisPJohn, line 63: There's a number of logs here, I assume those should get removed?

SolastaCommunityExpansion/Patches/CharacterExport/MessageModalPatcher.cs
ThyWoof: All game screens are loaded before main screen is displayed. The best place to change a screen is on the Load method but MessageModal doesn’t overwrite it. Setup a static that gets initialized only once at the very first time a message modal is displayed. From there it has this additional input field game object that gets activated or not depending on a special content.

SolastaCommunityExpansion/Models/CharacterExportContext.cs
ThyWoof: I’ll clean them up after @PhilPJL reviews the changes.

PhilPJL: Those logs only get used if you're exporting a character and it's a debug build, so don't pollute the log.

SolastaCommunityExpansion/Models/CharacterExportContext.cs
```diff
@@ -0,0 +1,236 @@
...
+
+                if (racePresentation.HasSurName && newFirstName.Contains(" "))
+                {
+                    var a = newFirstName.Split(new char[] { ' ' }, 2);
+
+                    newFirstName = a[0];
+                    newSurname = a[1];
+                }
+
+                if (!usedNames.Contains(newFirstName))
```
PhilPJL, line 128: If  `usedNames.Contains(newFirstName) == true ` we should ideally notify the user to try a different name (another message box or add a warning message to this message box) and prevent the modal from closing by patching OnValidateCb().  Maybe a future update?

SolastaCommunityExpansion/Patches/CharacterExport/RulesetInventoryPatcher.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace SolastaCommunityExpansion.Patches
+{
+    // When inspecting a character out of game, there is no RulesetEntityService registered.
+    // We need to temporarily add the dummy RulesetEntityService and then remove it.
+    // We don't want to register it or remove it when the real one has been registered.
+    [HarmonyPatch(typeof(RulesetInventory), "SerializeElements")]
+    internal static class RulesetInventory_SerializeElements
+    {
+        static readonly object Locker = new object();
```
PhilPJL, line 11: Not sure this locker stuff is required, but doesn't do any harm.  Was worried about threading when getting loads of null refs, but I don't think Unity does multi-threading.

PhilPJL (approved):

SolastaCommunityExpansion/Models/CharacterExportContext.cs
ThyWoof: that was considered but was lazy to think on a good solution. will fix.

SolastaCommunityExpansion/Patches/CharacterExport/RulesetInventoryPatcher.cs
ThyWoof: will leave it

ThyWoof:
took all the requests. closing this one to open a new request with latest changes in dev from this morning

PR #135, opened 2021-11-22
removes modkit reference as it's embed again in modapi

ThyWoof:


ThyWoof:
I've pushed latest 1.2.15 to ModApi beta. I tested both CE main against it without issues (except for the deprecated turn).

SolastaCommunityExpansion/SolastaCommunityExpansion.csproj
```diff
@@ -412,12 +408,4 @@
 		  <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
 		</None>
 	</ItemGroup>
-
```
PhilPJL, line 415: Where's editorconfig gone to?  Is it moved to solution level?

PhilPJL (approved):

PR #138, opened 2021-11-22
character export with changes from other request

ThyWoof:
. removed loggers
. added a message to inform hero with same name
. fixed a bug with classes without surnames


ThyWoof:
taking from other request...

SolastaCommunityExpansion/Models/CharacterExportContext.cs
```diff
@@ -0,0 +1,238 @@
...
+            {
+                string newFirstName = Patches.MessageModalPatcher.InputField.text;
+                string newSurname = string.Empty;
+                RacePresentation racePresentation = hero.RaceDefinition.RacePresentation;
+                HashSet<string> usedNames = Directory
+                    .EnumerateFiles(TacticalAdventuresApplication.GameCharactersDirectory, $"*.chr")
+                    .Select(f => Path.GetFileNameWithoutExtension(f))
+                    .ToHashSet(StringComparer.OrdinalIgnoreCase);
+
+                if (newFirstName.Contains(" "))
```
PhilPJL, line 108: Should we also check if(string.IsNullOrWhitespace(...)) { Alert }?

PhilPJL, line 108: Should it be` if(racePresentation.HasSurName && newFirstName.Contains(" "))`

PhilPJL (changes requested):
See comments

SolastaCommunityExpansion/Models/CharacterExportContext.cs
ThyWoof: good catch. will add this check.

PhilPJL (dismissed):

SolastaCommunityExpansion/Models/CharacterExportContext.cs
```diff
@@ -0,0 +1,255 @@
...
+                // TODO: remove weapon modifiers and effects
+
+                // remove attunement, attuned items don't work well in the character inspection screen out of game
+                foreach (var item in attunedItems)
+                {
+                    item.Item.AttunedToCharacter = string.Empty;
+                }
+
+                // clear guids
+                heroCharacter.SetGuid(0);
```
ChrisPJohn, line 191: I wonder if we should be calling heroCharacter.Unregister() and heroCharacter.ResetForOutgame() here?

This is based on HeroStatsColumn::OnUpdateCharacterCb

  public void OnUpdateCharacterCb()
  {
    this.UpdatedToDisk = true;
    this.updateFileButton.interactable = false;
    ICharacterPoolService service = ServiceRepository.GetService<ICharacterPoolService>();
    this.GuiCharacter.RulesetCharacterHero.Unregister();
    this.GuiCharacter.RulesetCharacterHero.ResetForOutgame();
    RulesetCharacterHero rulesetCharacterHero = this.GuiCharacter.RulesetCharacterHero;
    service.SaveCharacter(rulesetCharacterHero);
  }

ChrisPJohn (changes requested):
See comment about maybe calling unregister and resetforoutgame

SolastaCommunityExpansion/Models/CharacterExportContext.cs
ThyWoof: I only wrote the pop up code. @PhilPJL ?

SolastaCommunityExpansion/Models/CharacterExportContext.cs
```diff
@@ -0,0 +1,234 @@
...
+                {
+                    item.Item.SetGuid(0);
+                }
+
+                foreach (var item in inventoryItemGuids)
+                {
+                    item.Item.SetGuid(0);
+                }
+
+                heroCharacter.SetCurrentHitPoints(heroCharacter.CurrentHitPoints + heroCharacter.MissingHitPoints);
```
PhilPJL, line 189: This changes the in-game hp.  Need to save current hp and restore later on.

PhilPJL (changes requested):

SolastaCommunityExpansion/Models/CharacterExportContext.cs
PhilPJL: I did try those methods when they appeared thinking they would simplify things, but they don't because they're intended to work out of game.

SolastaCommunityExpansion/Models/CharacterExportContext.cs
```diff
@@ -0,0 +1,234 @@
...
+                heroCharacter.Unregister();
```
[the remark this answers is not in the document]
ThyWoof: cleaned up more code and now using Unregister() / Register() pair. Powers, Spells and Conditions still need to be saved manually.

ThyWoof:
@PhilPJL , trying to reuse as much as possible original TA code. It looks like we don't need to keep a tab on inventory items if we use Unregister / Register. Tested on a few saves.

ThyWoof:
Will remove commented code after we agree on a final version.

SolastaCommunityExpansion/Models/CharacterExportContext.cs
```diff
@@ -0,0 +1,238 @@
...
+            public ulong GenerateGuid()
+            {
+                return 0;
+            }
+
+            public void RegisterEntity(RulesetEntity rulesetEntity)
+            {
+                try
+                {
+                    if (rulesetEntity is RulesetItem)
```
PhilPJL, line 191: Remove these empty statements

PhilPJL (approved):
Looks good seems to work fine.  My only issue is that the text input needs a border or different colour background and needs to get the focus when the modal is opened.

ThyWoof:
made more changes but sometimes git gets on my way... closing this and will open a new one with latest dev commit.

PR #140, opened 2021-11-23
move init code off patch / final UI tweaks

ThyWoof:


SolastaCommunityExpansion/Patches/CharacterExport/MessageModalPatcher.cs
```diff
@@ -0,0 +1,30 @@
...
+                    InputField.text = string.Empty;
+                    ___contentLabel.TMP_Text.alignment = TextAlignmentOptions.BottomLeft;
+                }
+                else
+                {
+                    InputField.gameObject.SetActive(false);
+                }
+            }
+        }
+    }
```
PhilPJL, line 29: Adding this forces the focus to InputField

```
        [HarmonyPatch(typeof(MessageModal), "OnEndShow")]
        internal static class MessageModalOnEndShow
        {
            internal static void Postfix()
            {
                if (Main.Settings.EnableCharacterExport)
                {
                    EventSystem.current.SetSelectedGameObject(InputField.gameObject);
                    InputField.OnPointerClick(null);
                }
            }
        }
```

PhilPJL (changes requested):
Looks even better.  Input/modal not receiving focus though when shown.

SolastaCommunityExpansion/Patches/CharacterExport/MessageModalPatcher.cs
ThyWoof: Based on comment below should I consider this change?

ThyWoof: Added additional checks around InputField and a later instantiation call in case setting is enabled after Load.

SolastaCommunityExpansion/Models/CharacterExportContext.cs
```diff
@@ -0,0 +1,207 @@
...
+                // restore conditions
+                foreach (var kvp in conditions)
+                {
+                    heroCharacter.ConditionsByCategory.Add(kvp.Key, kvp.Value);
+                }
+
+                // restore active spells and effects
+                heroCharacter.PowersUsedByMe.AddRange(powers);
+                heroCharacter.SpellsCastByMe.AddRange(spells);
+                heroCharacter.Register(true);
```
ChrisPJohn, line 135: So it looks like this will end up with the hero having a different guid than before the export happened, is that a problem?

ThyWoof: less is more. brought back guid backup code to be safe.

ChrisPJohn (approved):
Make sure Phil approves as well. He's the export expert.

SolastaCommunityExpansion/Models/CharacterExportContext.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            // record current name, etc..
+            var guid = heroCharacter.Guid;
+            var firstName = heroCharacter.Name;
+            var surName = heroCharacter.SurName;
+            var builtin = heroCharacter.BuiltIn;
+            var hitPoints = heroCharacter.CurrentHitPoints;
+
+            // record current conditions, powers, spells and attunements
+            var conditions = heroCharacter.ConditionsByCategory.ToList();
```
PhilPJL, line 104: `conditions` is unused.  Need to restore conditions.

ThyWoof: fixed.

SolastaCommunityExpansion/Models/CharacterExportContext.cs
```diff
@@ -0,0 +1,233 @@
...
+
+                if (newFirstName == string.Empty)
+                {
+                    Gui.GuiService.ShowAlert("Message/&CharacterExportEmptyNameErrorDescription", "EA7171", 5);
+                }
+                else
+                {
+                    if (newFirstName.Contains(" "))
+                    {
+                        var a = newFirstName.Split(new char[] { ' ' }, 2);
```
PhilPJL, line 74: Please change to `.Split(' ', 2, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)`.
Otherwise things go wrong :)

PhilPJL: We also need to check for invalid filename characters as per the game and with the option to allow extra characters.

ThyWoof: added both validation types based on allow extra chars. setting.

PhilPJL: Ah ok, the parsetext is trimming.  Ok.

PhilPJL: You probably need TrimEntries at least

image

PhilPJL (approved):

PR #146, opened 2021-11-24
bring all patch classes under same namespace / standardize class names

ThyWoof:
. only class names changed
. all patches under same namespace to make it easier to identify bumps in the future (if we follow patch classes nomenclature)


PhilPJL (commented):
If all patches are under the same namespace
1) this doesn't conform to MS naming standards which say the namespace should match the folder structure, so we will get analyzer warnings (can be ignored).  
2) if we need to patch the same method for more than one reason Cheat/GameUI/... they will have the same class name and namespace and conflict.  It would in this case be desirable to conform to 1).
3) MS naming standards also say use Pascal case, no underscores.  Again not really bothered but analyzers complain.  For patches I like `<classname>_<methodname>`.

ThyWoof:
2. Is the main reason I’m checking this. I would rather know early on that other area in the mod is patching the same method.
3. Agree _ looks better when naming a patch class.

ThyWoof:
closing this one as I messed up with 1 merge...

PR #150, opened 2021-11-24
clean up patches to follow same name convention

ThyWoof:
Closed previous one and opened this. Continuing the discussion my justification to have all patch classes under *.Patches namespace is to guarantee we get an early error if trying to patch the same methods on a different feature.


ThyWoof:
will come back to this later... too much check to do

PR #160, opened 2021-11-28
Added Path of the Light barbarian subclass

RedOrcaCode:
- Path of the Light barbarian subclass.
- Generic ConditionDefinitionBuilder that supports any class that inherits from ConditionDefinition.
- Generic FeatureDefinitionBuilder that supports any class that inherits from FeatureDefinition.
- IClassHoldingFeature to allow IAdditionalDamageProvider features (like FeatureDefinitionAdditionalDamage) to scale with class level even if the feature isn't directly added to the class.
- IConditionRemovedOnSourceTurnStart to add support for conditions that last until the start of your next turn.
- INotifyConditionRemoval to support arbitrary behavior when a condition is removed (or the target dies).
- IStartOfTurnRecharge to add support for powers that recharge at the start of your next turn.

### DIFF
--- a/SolastaCommunityExpansion/Patches/ClassHoldingFeature/RulesetCharacterHeroPatcher.cs
+++ b/SolastaCommunityExpansion/Patches/ClassHoldingFeature/RulesetCharacterHeroPatcher.cs
@@ -1,0 +1,31 @@
+﻿using HarmonyLib;
+using SolastaCommunityExpansion.CustomFeatureDefinitions;
+
+namespace SolastaCommunityExpansion.Patches.ClassHoldingFeature
+{
+    internal static class RulesetCharacterHeroPatcher
+    {
+        [HarmonyPatch(typeof(RulesetCharacterHero), "FindClassHoldingFeature")]
+        internal static class RulesetCharacterHero_FindClassHoldingFeature
+        {
+            internal static void Postfix(
+                RulesetCharacterHero __instance,
+                FeatureDefinition featureDefinition,
+                ref CharacterClassDefinition __result)
+            {
+                var overrideClassHoldingFeature = featureDefinition as IClassHoldingFeature;
+
+                if (overrideClassHoldingFeature?.Class == null)
+                {
+                    return;
+                }
+
+                // Only override if the character actually has levels in the class, to prevent errors
+                if (__instance.ClassesAndLevels.TryGetValue(overrideClassHoldingFeature.Class, out int levelsInClass) && levelsInClass > 0)
+                {
+                    __result = overrideClassHoldingFeature.Class;
+                }
+            }
+        }
+    }
+}
--- a/SolastaCommunityExpansion/Subclasses/Barbarian/PathOfTheLight.cs
+++ b/SolastaCommunityExpansion/Subclasses/Barbarian/PathOfTheLight.cs
@@ -1,0 +1,966 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using SolastaCommunityExpansion.CustomFeatureDefinitions;
+using SolastaCommunityExpansion.Features;
+using SolastaModApi;
+using SolastaModApi.Extensions;
+using SolastaModApi.Infrastructure;
+
+namespace SolastaCommunityExpansion.Subclasses.Barbarian
+{
+    internal class PathOfTheLight : AbstractSubclass
+    {
+        private static readonly Guid SubclassNamespace = new Guid("c2067110-5086-45c0-b0c2-4c140599605c");
+        private const string IlluminatedConditionName = "PathOfTheLightIlluminatedCondition";
+        private const string IlluminatingStrikeName = "PathOfTheLightIlluminatingStrike";
+        private const string IlluminatingBurstName = "PathOfTheLightIlluminatingBurst";
+
+        private static readonly List<ConditionDefinition> InvisibleConditions =
+            new List<ConditionDefinition>
+            {
+                DatabaseHelper.ConditionDefinitions.ConditionInvisibleBase,
+                DatabaseHelper.ConditionDefinitions.ConditionInvisible,
+                DatabaseHelper.ConditionDefinitions.ConditionInvisibleGreater
+            };
+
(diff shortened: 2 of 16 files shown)


SolastaCommunityExpansion/Subclasses/Barbarian/PathOfTheLight.cs
```diff
@@ -0,0 +1,966 @@
...
+
+            // Intentionally *includes* conditions that have Illuminated as their parent (like the Illuminating Burst condition)
+            if (!character.HasConditionOfTypeOrSubType(IlluminatedConditionName))
+            {
+                if (character.PersonalLightSource?.SourceName == IlluminatingStrikeName ||
+                    character.PersonalLightSource?.SourceName == IlluminatingBurstName)
+                {
+                    var visibilityService = ServiceRepository.GetService<IGameLocationVisibilityService>();
+
+                    visibilityService.RemoveCharacterLightSource(GameLocationCharacter.GetFromActor(removedFrom), character.PersonalLightSource);
```
ChrisPJohn, line 503: Why do you need to manually remove the light sources here when Shine/Faerie Fire do not?

SolastaCommunityExpansion/Patches/ClassHoldingFeature/RulesetCharacterHeroPatcher.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        [HarmonyPatch(typeof(RulesetCharacterHero), "FindClassHoldingFeature")]
+        internal static class RulesetCharacterHero_FindClassHoldingFeature
+        {
+            internal static void Postfix(
+                RulesetCharacterHero __instance,
+                FeatureDefinition featureDefinition,
+                ref CharacterClassDefinition __result)
+            {
+                var overrideClassHoldingFeature = featureDefinition as IClassHoldingFeature;
```
ChrisPJohn, line 16: @ThyWoof I want your feedback on this new feature because you're the multiclass expert. In particular is this something you already handle properly?

SolastaCommunityExpansion/Subclasses/Barbarian/PathOfTheLight.cs
```diff
@@ -0,0 +1,966 @@
...
+                HandleAfterIlluminatedConditionRemoved(removedFrom);
+            }
+
+            public void BeforeDyingWithCondition(RulesetActor rulesetActor, RulesetCondition rulesetCondition)
+            {
+                ApplyLightsProtectionHealing(rulesetCondition.SourceGuid);
+            }
+        }
+
+        private class AttackDisadvantageAgainstNonCaster : FeatureDefinition, ICombatAffinityProvider
```
ChrisPJohn, line 538: This seems potentially useful in other circumstances. Would it make sense to add this (and maybe the OpportunityAttackImmunityIfAttackerHasCondition definition below as well) to CustomFeatureDefinitions?

ChrisPJohn (commented):
Several questions, which may turn into changes requested depending on the answers.

SolastaCommunityExpansion/Subclasses/Barbarian/PathOfTheLight.cs
```diff
@@ -0,0 +1,966 @@
...
+                return effectDescriptionBuilder.Build();
+            }
+        }
+
+        private class IlluminatingBurstPower : FeatureDefinitionPower, IStartOfTurnRecharge
+        {
+            public bool IsRechargeSilent => true;
+        }
+
+        private class IlluminatingBurstBuilder : BaseDefinitionBuilder<IlluminatingBurstPower>
```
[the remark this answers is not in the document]
RedOrcaCode: Unfortunately, OnRageStartChoice and OnRageStartAutomatic ignore whatever targeting you have set and only target Self. To get around that, I use OnRageStartAutomatic to remove the condition that is disabling Illuminating Burst (thus effectively enabling it).

SolastaCommunityExpansion/Subclasses/Barbarian/PathOfTheLight.cs
RedOrcaCode: I'm initiating Illuminating Strike's extra damage, condition, and light source automatically using FeatureDefinitionAdditionalDamage. If I recall correctly, that light source lasts forever, so I had to remove it manually. I tried putting the light source inside the condition (as one of its features), but that had no effect.

Shine and Faerie Fire, on the other hand, have an EffectDescription that can terminate their light sources at the appropriate time. (They also don't have to worry about lasting until the start of the initiator's turn, which is its own challenge.)

I tried an alternative solution using a power with ActivationTime.OnAttackHitAutomatic, which is now ActivationTime.OnAttackSpellHitAutomatic. It added a significant delay after each hit and didn't roll double dice on crit, so I abandoned it.

SolastaCommunityExpansion/Subclasses/Barbarian/PathOfTheLight.cs
RedOrcaCode: That's fine with me! I wasn't sure if they were too specific for general use or not.

ChrisPJohn: Yeah, I believe there's another subclass that can take advantage of the attack disadvantage against non caster feature, plus they can be examples to follow for someone else trying to do similar features. There may be a refactoring to make them more customizable (similar to how the FeatureDefinitionOnAttackHitEffect and the FeatureDefinitionConditionalPower take a delegate that specifies their custom behavior), but we can do that when there's a need for that functionality.

For now it'd be great if you could move these to their own files in the CustomFeatureDefinitions folder.

RedOrcaCode: Those two are now moved into CustomFeatureDefinitions. I renamed AttackDisadvantageAgainstNonCaster to AttackDisadvantageAgainstNonSource to avoid confusion about what "caster" means.

ChrisPJohn (approved):

PR #165, opened 2021-12-01
Replaces bug-like models with alternative visuals

DubhHerder:
adds toggle in Game UI sub-menu
identifies and replaces the 3d model, portrait and bestiary entry of 5 bug models that are used by a total of 20 monsters

### DIFF
--- a/SolastaCommunityExpansion/Viewers/Displays/GameUiDisplay.cs
+++ b/SolastaCommunityExpansion/Viewers/Displays/GameUiDisplay.cs
@@ -39,6 +39,12 @@ namespace SolastaCommunityExpansion.Viewers.Displays
                 Main.Settings.HideMonsterHitPoints = toggle;
             }
 
+            toggle = Main.Settings.RemoveBugVisualModels;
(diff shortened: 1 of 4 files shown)


ChrisPJohn (approved):

SolastaCommunityExpansion/Viewers/Displays/GameUiDisplay.cs
```diff
@@ -39,6 +39,12 @@ internal static void DisplayGameUi()
                 Main.Settings.HideMonsterHitPoints = toggle;
             }
 
+            toggle = Main.Settings.RemoveBugVisualModels;
+            if (UI.Toggle("Replaces bug-like models with alternative visuals in the game" + RequiresRestart, ref toggle, UI.AutoWidth()))
+            {
+                Main.Settings.RemoveBugVisualModels = toggle;
```
ChrisPJohn, line 45: Assuming the game hasn't cached anything you could call load from here to immediately swap bugs -> animals.

DubhHerder: I have added the call to load here as suggested in the latest commit

ChrisPJohn (approved):

PR #171, opened 2021-12-05
Fix 'hide monster hit points affects wild-shaped druid'

PhilPJL:
This will close https://github.com/ChrisPJohn/SolastaCommunityExpansion/issues/170
Please check my 'IsMonster' logic.

### DIFF
--- a/SolastaCommunityExpansion/Patches/GameUi/HideMonsterHitPoints.cs
+++ b/SolastaCommunityExpansion/Patches/GameUi/HideMonsterHitPoints.cs
@@ -153,7 +153,7 @@ namespace SolastaCommunityExpansion.Patches.GameUi
                     return false;
                 }
 
-                if (__instance.RulesetCharacter?.IsSubstitute == true)
+                if (__instance.RulesetCharacter.IsSubstitute)
                 {
                     // It's a hero wildshaping (probably).
                     return false;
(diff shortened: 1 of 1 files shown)


SolastaCommunityExpansion/Patches/GameUi/HideMonsterHitPoints.cs
```diff
@@ -140,27 +144,45 @@ internal static void Postfix(GuiCharacter __instance, GuiLabel healthLabel, bool
...
             {
-                if (!Main.Settings.HideMonsterHitPoints) return;
+                if (__instance.RulesetCharacterMonster == null)
+                {
+                    // definitely not a monster
+                    return false;
+                }
 
-                if (__instance.GuiCharacter.RulesetCharacterMonster != null) // Only change for monsters
+                if (__instance.RulesetCharacter?.IsSubstitute == true)
```
ThyWoof, line 156: I believe RulesetCharacter will never be null here.

PhilPJL: you're right.  It's RulesetCharacterHero that is set to null.

PhilPJL: Fixed

ThyWoof (approved):

PR #179, opened 2021-12-06
uses FormatTitle / FormatDescription instead of Gui.Format

ThyWoof:


ChrisPJohn (approved):

SolastaCommunityExpansion/Models/FeatsContext.cs
```diff
@@ -38,7 +38,7 @@ internal static void Load()
                 feat.GuiPresentation.SetHidden(!Main.Settings.FeatEnabled.Contains(feat.Name));
             }
 
-            Feats = Feats.OrderBy(x => Gui.Format(x.Value.GuiPresentation.Title)).ToDictionary(x => x.Key, x => x.Value);
+            Feats = Feats.OrderBy(x => x.Value.FormatTitle()).ToDictionary(x => x.Key, x => x.Value);
```
PhilPJL, line 41: I never knew dictionary keys were stored in the order inserted.  Live and learn :)

PhilPJL (approved):

PR #189, opened 2021-12-07
Combat surprise

ThyWoof:
. Implements surprise rules much closer to SRD:

for each surprised entity:
  for each surprising entity in field of vision:
    roll surprising entity stealth check against passive surprised entity perception check
if at least one roll fails surprised entity isn't surprised anymore

### DIFF
--- a/SolastaCommunityExpansion/Models/SrdAndHouseRulesContext.cs
+++ b/SolastaCommunityExpansion/Models/SrdAndHouseRulesContext.cs
@@ -55,13 +55,22 @@ namespace SolastaCommunityExpansion.Models
                     {
                         if (gameLocationBattleService.CanAttackerSeeCharacterFromPosition(surprisingCharacter.LocationPosition, surprisedCharacter.LocationPosition, surprisingCharacter, surprisedCharacter))
                         {
-                            int perceptionOnTarget = surprisedCharacter.ComputePassivePerceptionOnTarget(surprisingCharacter, out bool _);
+                            int perceptionOnTarget = SURPRISE_CHECK_DC_BASE;
+                            
+                            if (surprisedCharacter.RulesetCharacter is RulesetCharacterMonster monster)
+                            {
+                                perceptionOnTarget = monster.MonsterDefinition.ComputePassivePerceptionScore();
+                            }
+                            else if (surprisedCharacter.RulesetCharacter is RulesetCharacterHero hero)
+                            {
+                                perceptionOnTarget = hero.ComputePassivePerception();
+                            }
 
-                            surprisingCharacter.RollAbilityCheck("Dexterity", "Stealth", SURPRISE_CHECK_DC_BASE + perceptionOnTarget, RuleDefinitions.AdvantageType.None, new ActionModifier(), false, -1, out RuleDefinitions.RollOutcome outcome, true);
+                            surprisingCharacter.RollAbilityCheck("Dexterity", "Stealth", perceptionOnTarget, RuleDefinitions.AdvantageType.None, new ActionModifier(), false, -1, out RuleDefinitions.RollOutcome outcome, true);
 
                             if (outcome == RuleDefinitions.RollOutcome.CriticalFailure || outcome == RuleDefinitions.RollOutcome.Failure)
                             {
-                                var conditionUnsurprised = UnityEngine.Object.Instantiate(DatabaseRepository.GetDatabase<ConditionDefinition>().GetElement("ConditionUnsurprised"));
+                                var conditionUnsurprised = DatabaseRepository.GetDatabase<ConditionDefinition>().GetElement("ConditionUnsurprised");
 
                                 surprisedCharacter.RulesetCharacter.AddConditionOfCategory("10Combat", RulesetCondition.CreateActiveCondition(surprisedCharacter.RulesetCharacter.Guid, conditionUnsurprised, RuleDefinitions.DurationType.Round, 0, RuleDefinitions.TurnOccurenceType.EndOfTurn, 0UL, string.Empty));
                                 isReallySurprised = false;
--- a/SolastaCommunityExpansion/Models/SrdAndHouseRulesContext.cs
+++ b/SolastaCommunityExpansion/Models/SrdAndHouseRulesContext.cs
@@ -10,15 +10,15 @@ namespace SolastaCommunityExpansion.Models
         internal static void Load()
         {
             var dbConditionDefinition = DatabaseRepository.GetDatabase<ConditionDefinition>();
-            var conditionUnsurprised = UnityEngine.Object.Instantiate(dbConditionDefinition.GetElement("ConditionSurprised"));
+            var conditionAware = UnityEngine.Object.Instantiate(dbConditionDefinition.GetElement("ConditionSurprised"));
 
-            conditionUnsurprised.name = "ConditionUnsurprised";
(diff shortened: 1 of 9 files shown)


ThyWoof:
image
here both heroes succeed on the stealth check so enemy got surprised.

ThyWoof:
image
here was playing with 4 heroes and second one failed. remaining heroes didn't roll stealth and enemy wasn't surprised anymore

ThyWoof:
Note that different than TT rules it will roll a new stealth check for each (surprised, surprising) tuples. It's a computer...

SolastaCommunityExpansion/Patches/SrdAndHouseRulesContext/GameLocationBattlePatcher.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            return true;
+        }
+
+        internal static void StartContenders(bool surprised, List<GameLocationCharacter> surprisedParty, List<GameLocationCharacter> surprisingParty)
+        {
+            var gameLocationBattleService = ServiceRepository.GetService<IGameLocationBattleService>();
+
+            if (surprised)
```
ChrisPJohn, line 28: You should start the method with an early exit instead of this if:

if(!surprised) {
return;
}

That reduces the indentation of everything.

SolastaCommunityExpansion/Patches/SrdAndHouseRulesContext/GameLocationBattlePatcher.cs
```diff
@@ -0,0 +1,63 @@
...
+                {
+                    var isReallySurprised = true;
+
+                    foreach (GameLocationCharacter surprisingCharacter in surprisingParty)
+                    {
+                        if (gameLocationBattleService.CanAttackerSeeCharacterFromPosition(surprisingCharacter.LocationPosition, surprisedCharacter.LocationPosition, surprisingCharacter, surprisedCharacter))
+                        {
+                            int perceptionOnTarget = surprisedCharacter.ComputePassivePerceptionOnTarget(surprisingCharacter, out bool _);
+
+                            surprisingCharacter.RollAbilityCheck("Dexterity", "Stealth", perceptionOnTarget, RuleDefinitions.AdvantageType.None, new ActionModifier(), false, -1, out RuleDefinitions.RollOutcome outcome, true);
```
ChrisPJohn, line 40: I saw there were comments in the discord about using an opposed check instead of the passive perception. Do we want to make that change?

My thought is we can iterate over the surprising characters and roll stealth checks, then iterate over the surprised characters and roll perception checks. If the surprised character rolls below all of the stealth checks of characters they can see, they're surprised, otherwise that character is not surprised.

This would mean we don't do nested foreach and I think gets it closer to the table top rules. The only issue is needing to specify a dc for the skill check.

The other possibility is to use ResolveContestCheck, though then we're back to a contested check for each pair.

ChrisPJohn (changes requested):
Please switch to using an early exit instead of an if around nearly the entire method.

Also I added some thoughts on how to make it closer to tabeltop rules.

SolastaCommunityExpansion/Patches/SrdAndHouseRulesContext/GameLocationBattlePatcher.cs
ThyWoof: not here. It's an if-else...

SolastaCommunityExpansion/Patches/SrdAndHouseRulesContext/GameLocationBattlePatcher.cs
ThyWoof: I would rather have it as is and get further feedback from players...

SolastaCommunityExpansion/Patches/SrdAndHouseRulesContext/GameLocationBattlePatcher.cs
ThyWoof: Still refactored code a bit to only have one outer loop

ThyWoof:
added a new dummy unsurprised condition so it clear reflects on console. FYI @ChrisPJohn

PhilPJL (dismissed):

ThyWoof:
we had a chance to test this and fix a bug on passive perception calc on monsters...

SolastaCommunityExpansion/Models/SrdAndHouseRulesContext.cs
```diff
@@ -1,11 +1,27 @@
...
 
 namespace SolastaCommunityExpansion.Models
 {
     internal static class SrdAndHouseRulesContext
     {
-        public static void ApplyConditionBlindedShouldNotAllowOpportunityAttack()
+        internal static void Load()
+        {
+            var dbConditionDefinition = DatabaseRepository.GetDatabase<ConditionDefinition>();
+            var conditionAware = UnityEngine.Object.Instantiate(dbConditionDefinition.GetElement("ConditionSurprised"));
```
ChrisPJohn, line 13: Is there a reason not to use a conditiondefinition builder here?

SolastaCommunityExpansion/Patches/SrdAndHouseRules/GameLocationBattlePatcher.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    [HarmonyPatch(typeof(GameLocationBattle), "StartContenders")]
+    [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
+    internal static class GameLocationBattle_StartContenders
+    {
+        internal static bool Prefix(GameLocationBattle __instance, bool partySurprised, bool enemySurprised)
+        {
+            if (Main.Settings.EnableSRDCombatSurpriseRules && (partySurprised || enemySurprised))
+            {
+                Models.SrdAndHouseRulesContext.StartContenders(partySurprised, __instance.PlayerContenders, __instance.EnemyContenders);
```
ChrisPJohn, line 14: Is it possible for both groups to be surprised or do we really only need to call one of these 2 each time?

ChrisPJohn (changes requested):
See comments

SolastaCommunityExpansion/Patches/SrdAndHouseRules/GameLocationBattlePatcher.cs
ThyWoof: both cannot be surprised. still need to call both as all contenders must be initialized. when our StartContender gets called with surprised = false it will default to standard game output (the inner for loop never executes and contender is started as not surprised).

SolastaCommunityExpansion/Models/SrdAndHouseRulesContext.cs
```diff
@@ -23,5 +39,49 @@ public static void ApplyConditionBlindedShouldNotAllowOpportunityAttack()
...
+
+        internal static void StartContenders(bool surprised, List<GameLocationCharacter> surprisedParty, List<GameLocationCharacter> surprisingParty)
+        {
+            var gameLocationBattleService = ServiceRepository.GetService<IGameLocationBattleService>();
+
+            foreach (GameLocationCharacter surprisedCharacter in surprisedParty)
+            {
+                var isReallySurprised = true;
+
+                if (surprised)
```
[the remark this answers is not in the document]
ThyWoof: no. trust me. this if as is is required... otherwise a contender won't be added to battle.

ThyWoof:
closing this as refactored Game Console touching less files

PR #190, opened 2021-12-07
remove side effects from Settings setters

PhilPJL:
Applying settings during the xml deserialization was throwing exception after recent change to make some static properties lazy loading.
Now applying at the end of GameManager_BindPostDatabase_Patch.

### DIFF
--- a/SolastaCommunityExpansion/Patches/GameManagerPatcher.cs
+++ b/SolastaCommunityExpansion/Patches/GameManagerPatcher.cs
@@ -1,5 +1,9 @@
 ﻿using System.Diagnostics.CodeAnalysis;
 using HarmonyLib;
+using SolastaCommunityExpansion.Feats;
+using SolastaCommunityExpansion.Models;
+using SolastaCommunityExpansion.Subclasses.Rogue;
+using SolastaCommunityExpansion.Subclasses.Wizard;
 
 namespace SolastaCommunityExpansion.Patches
 {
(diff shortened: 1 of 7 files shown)


SolastaCommunityExpansion/Patches/GameManagerPatcher.cs
```diff
@@ -11,30 +15,36 @@ internal static class GameManager_BindPostDatabase_Patch
...
+                RespecContext.Load();
+                RemoveIdentificationContext.Load();
+                Level20Context.Load();
+                DruidArmorContext.Load();
+                CharacterExportContext.Load();
+                InventoryManagementContext.Load();
+                RemoveBugVisualModelsContext.Load();
+
+                // Apply settings after everything loaded
+                ConArtist.UpdateSpellDCBoost();
```
ChrisPJohn, line 44: I believe most or all of these apply the currently loaded settings as part of the load call. Do they need to be called from here?

SolastaCommunityExpansion/Settings.cs
```diff
@@ -57,15 +56,7 @@ public class Settings : UnityModManager.ModSettings
         public List<string> ItemsInDM { get; private set; } = new List<string>();
         public List<string> RecipesInDM { get; private set; } = new List<string>();
 
-        private int recipeCost = 200;
-        public int RecipeCost
-        {
-            get => recipeCost; set
-            {
-                recipeCost = value;
-                Models.ItemCraftingContext.UpdateRecipeCost();
```
ChrisPJohn, line 66: The calls as part of the setter are what make it so these settings don't require restarts. If you're going to remove the calls from here then you need to add them to the display when the settings are changed.

ChrisPJohn (changes requested):
Add calls in the display code to update the cost, subclasses, and feats when the settings are changed.

SolastaCommunityExpansion/Settings.cs
PhilPJL: Ok, I see.  I can either reinstate like this

```
            get => recipeCost; set
            {
                recipeCost = value;
                if(Main.Enabled){  Models.ItemCraftingContext.UpdateRecipeCost(); }
```

Or as you say

```
            if (UI.Slider("Recipes' Cost".white(), ref intValue, 1, 500, 200, "", UI.AutoWidth()))
            {
                Main.Settings.RecipeCost = intValue;
                ItemCraftingContext.UpdateRecipeCost();
            }

```

I prefer the 2nd approach since Settings is effectively a DTO and shouldn't have side effects like that.

SolastaCommunityExpansion/Viewers/Displays/FeatsDisplay.cs
```diff
@@ -24,6 +25,7 @@ internal static void DisplayFeats()
             if (UI.Slider("Power Attack modifier ".white() + RequiresRestart, ref intValue, 1, 6, 3, ""))
```
PhilPJL, line 25: Does this 'require restart'?

ThyWoof: It does unless we update the gui presentation texts on a change

ThyWoof (approved):

ChrisPJohn (approved):

PR #201, opened 2021-12-11
fix user campaigns folders

ThyWoof:


PhilPJL (commented):
without testing, does that need to be 

string.IsNullOrEmpty/Whitespace(session.UserCampaign) ? ... : ...

PhilPJL (approved):
Otherwise fine.

PhilPJL (approved):

PR #206, opened 2021-12-11
Witch class

RealBazou:
I cannot stress how grateful I am to you guys. I've never attempted something like this, so I am really out of my comfort zone.

I still have a lot of questions, and probably will for a while. I have decided to limit my scope to what is in this branch for now. There is just too much to do, and hopefully this would be a nice Merry Christmas present to the community.

How should I proceed from here on out? I have tested this "organically", i.e. as I was developing features. I would have loved to be able to toggle on/off the class itself, but I am not sure how to make it disappear from the list of available classes in the character creation screen. I invite everyone to give a go at it if you have some time. 

Here are notes I have taken, as well as more notes in the comments in the code:

FOR CE REPO ADMINS:

- Added a menu for global powers (Help Action)
- Added a menu for global spells (Eldritch Orb, Find Familiar, Frenzy, Minor Lifesteal, Petal Storm, Protect Threshold)
- NO added menu for classes, as I do not know how to make the class option disappear from the class selection menu during character creation or level up
- Tried to use API as much as I could, but sometimes copying from existing DB items was simpler and some methods weren't always available
- Did some patching, and one prefix patch for RulesetImplementationManagerLocation.ApplySummonForm with a return false
- Sometimes I check for DB presence of a spell or class, however I am not sure if this is the best way to do things. It's probably better to check for the presence of the item in the enabled List? I will refer to your judgment here in regards to mod compatibility and player experience. For now I check DB presence.

SPELLS

- Eldritch Orb: known issue: only main hit should do 1d8 to primary target, and the aoe should do half to surrounding targets. Right now it applies both to everyone.
- Find Familiar: Owl form only, other familiar forms not implemented
- Find Familiar: dismiss and reappear within 30 feet as an action is not implemented
- Find Familiar: unsummon is not implemented
- Find Familiar: remote casting of touch spell while within 100 feet using a reaction is not implemented
- Frenzy: known issue: the frenzied effect doesn't work when enemies are not adjacent to a target. This is also true of the stock Confusion spell it seems.
- Minor Lifesteal: known issue: should give back the damage as temporary hit points, right now it heals (Vampiric Touch effect).
- Petal Storm: known issue: no added damage when spell is above difficult terrain

WITCH CURSES

- Burned: known issue: Unable to give Produce Flame cantrip at level 1.
- Visions: known issue: should be Charisma based, Flat +3 bonus for now

MALEDICTIONS

- 2nd concentration mechanic is not implemented

WITCH FAMILIAR

- Owl form only, other familiar forms not implemented
- Dismiss and reappear within 30 feet as an action is not implemented
- Unsummon is not implemented
- Remote casting of any spell while within 100 feet using a reaction is not implemented
- Same initiative as caster is not implemented
- Reaction to attack, triggered by caster's action or bonus action is not implemented. Right now, witch familiar attacks freely.

CACKLE

- not implemented, only gives a Deafened condition for now.

INSIDIOUS SPELL

- not implemented

IMPROVED FAMILIAR

- not implemented

DYING CURSE

- not implemented

GRAND MALEDICTIONS

- not implemented

ABSOLUTE MALEDICTION

- not implemented

SUBCLASSES

- Green Witch: only Auto-prepared spells are implemented
- Red Witch: only Auto-prepared spells are implemented
- White Witch: only Auto-prepared spells are implemented


SolastaCommunityExpansion/Features/FeatureDefinitionFeatureSetBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        public FeatureDefinitionFeatureSetBuilder ClearFeatures()
+        {
+            Definition.FeatureSet.Clear();
+            return this;
+        }        public FeatureDefinitionFeatureSetBuilder SetFeature(FeatureDefinition featureDefinition)
+        {
+            
+            // IMPROVE THIS AND SEARCH FOR FEATURESET TYPE... THERE IS NO SETTER FOR FEATURESET
```
RealBazou, line 28: This is terrible, I know. How do we do this cleanly?

RedOrcaCode: I recently added a generic FeatureDefinitionBuilder that accepts any FeatureDefinition (including FeatureDefinitionFeatureSet). That way, we don't have to make a new builder for every FeatureDefinition we end up using.

Here's an example of me using it to make a FeatureDefinitionFeatureSet for Path of the Light:

https://github.com/ChrisPJohn/SolastaCommunityExpansion/blob/53139a743ccde604d0486069b84264c7f807af23/SolastaCommunityExpansion/Subclasses/Barbarian/PathOfTheLight.cs#L90-L117

The important part is using SetField to initialize the list of features. I wouldn't exactly call that "clean", but it's our best option at the moment.

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -0,0 +1,1277 @@
...
+                Definition.SetSpecificMaterialComponentConsumed(true);
+                Definition.SetSpecificMaterialComponentCostGp(10);
+                Definition.SetCastingTime(RuleDefinitions.ActivationTime.Hours1);
+                Definition.SetRitual(true);
+
+                // BUG: Unable to have 70 minutes ritual casting time... if set to 10 minutes, it really only takes 10 minutes, isntead of 70
+                Definition.SetRitualCastingTime(RuleDefinitions.ActivationTime.Hours1);
+                Definition.SetRequiresConcentration(false);
+
+                var effectForm = new EffectForm();
```
ChrisPJohn, line 895: FYI There is an EffectFormBuilder in the SolastaModApi which could help here.

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -0,0 +1,1277 @@
...
+                SummonForm summonForm = new SummonForm();
+                effectForm.SetSummonForm(summonForm);
+
+                MonsterDefinition monsterDefinition = new MonsterDefinition();
+                monsterDefinition = FamiliarMonsterBuilder.AddToMonsterList();
+
+                effectForm.SummonForm.SetMonsterDefinitionName(monsterDefinition.Name);
+                effectForm.SummonForm.SetDecisionPackage(null);
+
+                Definition.EffectDescription.Copy(DatabaseHelper.SpellDefinitions.ConjureAnimalsOneBeast.EffectDescription);
```
ChrisPJohn, line 908: FYI There is a EffectDescriptionBuilder in the SolastaModApi which could help here.

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -0,0 +1,1277 @@
...
+                                                                                    new GuiPresentationBuilder(
+                                                                                        "Class/&WitchFamiliarPowerDescription",
+                                                                                        "Class/&WitchFamiliarPowerTitle").Build(),
+                                                                                    true)
+                                                                                    .AddToDB();
+
+            return familiar;
+        }
+
+        internal class FindFamiliarSpellBuilder : BaseDefinitionBuilder<SpellDefinition>
```
ChrisPJohn, line 870: FYI There is a SpellBuilder in the SolastaModApi which could help here.

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -0,0 +1,1277 @@
...
+
+                Definition.SetSpellLevel(1);
+                Definition.SetSchoolOfMagic(RuleDefinitions.SchoolConjuration);
+                Definition.SetVerboseComponent(true);
+                Definition.SetSomaticComponent(true);
+                Definition.SetMaterialComponentType(RuleDefinitions.MaterialComponentType.Specific);
+                Definition.SetSpecificMaterialComponentConsumed(true);
+                Definition.SetSpecificMaterialComponentCostGp(10);
+                Definition.SetCastingTime(RuleDefinitions.ActivationTime.Hours1);
+                Definition.SetRitual(true);
```
ChrisPJohn, line 889: You mentioned being able to summon infinite familiars. You probably want to use:

Definition.SetUniqueInstance(true);

I've written very few spells, but the similar setting for powers does what you want.

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -0,0 +1,1277 @@
...
+
+            witch.AddSkillPreference(DatabaseHelper.SkillDefinitions.Arcana);
+            witch.AddSkillPreference(DatabaseHelper.SkillDefinitions.Deception);
+            witch.AddSkillPreference(DatabaseHelper.SkillDefinitions.Insight);
+            witch.AddSkillPreference(DatabaseHelper.SkillDefinitions.Intimidation);
+            witch.AddSkillPreference(DatabaseHelper.SkillDefinitions.Persuasion);
+            witch.AddSkillPreference(DatabaseHelper.SkillDefinitions.Nature);
+            witch.AddSkillPreference(DatabaseHelper.SkillDefinitions.Religion);
+            
+            witch.AddEquipmentRow(  new List<CharacterClassDefinition.HeroEquipmentOption>
```
ChrisPJohn, line 59: You mentioned wanting to use the dropdowns for equipment selection. I think you want something like:

```
List<HeroEquipmentOption> simpleWeaponList = new List<HeroEquipmentOption>();
            HeroEquipmentOption simpleWeaponOption = EquipmentOptionsBuilder.Option(DatabaseHelper.ItemDefinitions.Mace, EquipmentDefinitions.OptionWeaponSimpleChoice, 1);
            simpleWeaponList.Add(simpleWeaponOption);
            simpleWeaponList.Add(simpleWeaponOption);
            ArtificerBuilder.AddEquipmentRow(simpleWeaponList);

List<HeroEquipmentOption> focus = new List<HeroEquipmentOption>
            {
                EquipmentOptionsBuilder.Option(DatabaseHelper.ItemDefinitions.ArcaneFocusWand, EquipmentDefinitions.OptionArcaneFocusChoice, 1)
            };
            ArtificerBuilder.AddEquipmentRow(focus);
```

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -0,0 +1,1277 @@
...
+//            witch.AddFeatureAtLevel(Cackle,2);
+
+            witch.AddFeatureAtLevel(DatabaseHelper.FeatureDefinitionFeatureSets.FeatureSetAbilityScoreChoice, 4);
+
+            // TODO: Maledictions should now apply a debuff for disadvantage on saving throw like Force Of Law
+//            witch.AddFeatureAtLevel(InsidiousSpell,5);
+
+            witch.AddFeatureAtLevel(maledictions, 5);
+
+            // TODO: Simply buff the familiar accordingly, i.e. offer more forms, and if that is too hard,
```
ChrisPJohn, line 446: I don't think we have a way to change the effects of a spell based on caster level currently, but theoretically that's a feature we could add.

That said, it is possible to have summoned creatures advance with the summoner level. This was added for the Kindred Spirits Druid. @DubhHerder is more familiar with that than me and might have some useful advice.

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -0,0 +1,1277 @@
...
+            // This should actually be a bit refined, i.e. give the Find Familiar spell, and have a longer casting time/ritual tag
+            // Beckon Familiar is actually a Malediction, i.e. instant cast familiar
+            witch.AddFeatureAtLevel(Familiar(spellList), 2);
+            // TODO: unsure about this one...
+            // Needs to refresh an existing Malediction, preventing the saving throw again I presume
+//            witch.AddFeatureAtLevel(Cackle,2);
+
+            witch.AddFeatureAtLevel(DatabaseHelper.FeatureDefinitionFeatureSets.FeatureSetAbilityScoreChoice, 4);
+
+            // TODO: Maledictions should now apply a debuff for disadvantage on saving throw like Force Of Law
```
ChrisPJohn, line 441: I don't know if this is what you're looking for, but possibly the new custom feature to have additional effects on attack hit would be of use:

https://github.com/ChrisPJohn/SolastaCommunityExpansion/blob/Dev/SolastaCommunityExpansion/CustomFeatureDefinitions/FeatureDefinitionOnAttackHitEffect.cs#L25

You can see it used:
https://github.com/ChrisPJohn/SolastaCommunityExpansion/blob/Dev/SolastaCommunityExpansion/Feats/ElAntoniousFeats.cs#L43

I don't know if this supports saving throw based effects, I'd have to go dig through the game code to check. We can probably do something similar for those if needed.

SolastaCommunityExpansion/Classes/Witch.cs
DubhHerder: There is a limited way to change the effects of a spell based on caster level but it only has a set list of options designed to change the direct effect rather than change the buff a summons so it's not really applicable here as Chris says.

FeatureDefinitionSummoningAffinity is the method to buff summons with with PC level, i'm not sure if its class level or character level by applying conditions to summons with a specified tag. the conditions are used to apply features that could be normal features or that could use the summoners stats to determine the buff. Theres no option to increase via proficiency bonus, the summoner based options are   SourceHalfHitPoints,  SourceSpellCastingAbility, SourceSpellAttack.

so you could have the normal spell to summon a creature at low levels then later add a FeatureDefinitionSummoningAffinity to the summoner class at a higher level that grants an extra attack or improve the creatures to hit, damage, ac, hp etc based on the above, provided the summons has specified tag etc

SolastaCommunityExpansion/Models/SubclassesContext.cs
```diff
@@ -26,6 +27,9 @@ internal static void Load()
             LoadSubclass(new Tactician());
             LoadSubclass(new RoyalKnight());
             LoadSubclass(new PathOfTheLight());
+            LoadSubclass(new BloodWitch());
```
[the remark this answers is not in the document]
DubhHerder: i agree that is likely to be confusing and might result in false bug reports. loading subclasses related to a new class dont need to follow the same template as loading subclasses related to existing classes

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -0,0 +1,1277 @@
...
+            // lvl 9:  5
+            // lvl 13: 6
+            // lvl 17: 7
+
+            FeatureDefinitionFeatureSet maledictions = Maledictions();
+
+            witch.AddFeatureAtLevel(maledictions, 1);
+            witch.AddFeatureAtLevel(maledictions, 1);
+
+            // BUG: this will not work if applied at level 1, since not all data structures are built before choosing features.
```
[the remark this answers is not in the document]
RealBazou: It was failing to apply, so perhaps I was doing something wrong. I'll try to see maybe how Clerics do it?

SolastaCommunityExpansion/Classes/Witch.cs
RealBazou: I have fixed this indeed by adding 
                Definition.SetUniqueInstance(true);

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -472,6 +461,9 @@ internal Witch()
 
             // add class to db
             Class = witch.AddToDB();
+
+            // How to add an auto-prepared spell if the class has not been instanciated? the caster definition doesn't exist yet
```
ChrisPJohn, line 465: You should be able to add prepared spells. I've done it in Tinkerer. Here's an example in one of the CE subclasses:

https://github.com/ChrisPJohn/SolastaCommunityExpansion/blob/Dev/SolastaCommunityExpansion/Subclasses/Ranger/Arcanist.cs#L71-L101

Note you don't need to use a feature set, you should just be able to do `witch.AddFeatureAtLevel(preparedSpells, 2);`

RealBazou: I'm sorry, maybe I don't understand well. Apologies if this is the case.
I see the Arcanist Ranger uses DatabaseHelper.CharacterClassDefinitions.Ranger for its ClassDefinition, but in the case of a brand new class like Witch, how can we use the .SetCharacterClass function from the builder if this setup is being done while we are building up the class and before it has been added in the db/instanciated?

ThyWoof: check https://github.com/ThyWoof/SolastaUnfinishedBusiness/blob/master/SolastaUnfinishedBusiness/ClassWarlock/ClassWarlockBuilder.cs and try something similar... @ChrisPJohn and @DubhHerder are way more knowledgeable in this area. All I did was copy and paste some of their code there.

ThyWoof: check https://github.com/ThyWoof/SolastaUnfinishedBusiness/blob/master/SolastaUnfinishedBusiness/ClassWarlock/ClassWarlockBuilder.cs and try something similar... @ChrisPJohn and @DubhHerder are way more knowledgeable in this area. All I did was copy and paste some of their code there.

SolastaCommunityExpansion/Features/FeatureDefinitionFeatureSetBuilder.cs
RealBazou: Got rid of this permanently. Proceeded to clear and add.

SolastaCommunityExpansion/Models/SubclassesContext.cs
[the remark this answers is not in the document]
RealBazou: Removed all witch subclasses from the subclass settings screen.

SolastaCommunityExpansion/Patches/Cheats/RulesetCharacterHeroPatcher.cs
```diff
@@ -78,4 +80,40 @@ internal static void Postfix(ref int __result)
...
+
+    [HarmonyPatch(typeof(RulesetCharacterHero), "EnumerateUsableRitualSpells")]
+    internal static class RestModuleHitDice_EnumerateUsableRitualSpells_Patch
+    {
+        internal static void Postfix(RulesetCharacterHero __instance, RuleDefinitions.RitualCasting ritualType, List<SpellDefinition> ritualSpells)
+        {
+            var extended_ritual_type = (ExtraRitualCasting)ritualType;
+            switch (extended_ritual_type)
+            {
+                case ExtraRitualCasting.Known:
```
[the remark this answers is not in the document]
RealBazou: This was changed by Phil AFAIK. Thank you Phil.

SolastaCommunityExpansion/Level20/Classes/WitchBuilder.cs
```diff
@@ -0,0 +1,42 @@
+/*using SolastaCommunityExpansion.Level20.Features;
```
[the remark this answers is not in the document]
RealBazou: Removed the useless file. Now everything is in the witch class.

SolastaCommunityExpansion/Models/ClassesContext.cs
```diff
@@ -0,0 +1,42 @@
+﻿using SolastaCommunityExpansion.Classes;
+using SolastaCommunityExpansion.Classes.Witch;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace SolastaCommunityExpansion.Models
+{
+    internal static class ClassesContext
+    {
+        public static Dictionary<string, AbstractClass> Classes = new Dictionary<string, AbstractClass>();
```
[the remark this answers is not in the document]
RealBazou: I kept it as is for now, will defer to everyone if this is to be changed or not.

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -0,0 +1,1277 @@
...
+            witch.AddFeatureAtLevel(savingThrowProficiencies, 1);
+            witch.AddFeatureAtLevel(skillProficiencies, 1);
+            witch.AddFeatureAtLevel(weaponProficiencies, 1);
+            witch.AddFeatureAtLevel(armorProficiencies, 1);
+            witch.AddFeatureAtLevel(toolProficiencies, 1);
+
+            witch.AddFeatureAtLevel(spellCasting.AddToDB(), 1);
+            witch.AddFeatureAtLevel(ritualCasting, 1);
+
+            // TODO: Do some powers like the Tinkerer infusions, dropdown like infusions too
```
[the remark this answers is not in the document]
RealBazou: I am now using FeatureSets whenevr I need them. SharedPool is not what I am looking for (Witch powers do not have a limited amount), but thanks for the tip!

SolastaCommunityExpansion/Classes/Witch.cs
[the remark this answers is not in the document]
RealBazou: So this one is still outstanding for me, and I will simply leave it in the pile of known issues for now.

SolastaCommunityExpansion/Classes/Witch.cs
RealBazou: I have started using the API for this, but I find that sometimes it's easier for me to work from a copy that I want to edit surgically. The API sadly doesn't cover every attribute. I'm trying to use them as much as I can without creating overhead.

SolastaCommunityExpansion/Classes/Witch.cs
RealBazou: Same as above, I have started using the API for this, but I find that sometimes it's easier for me to work from a copy that I want to edit surgically. The API sadly doesn't cover every attribute. I'm trying to use them as much as I can without creating overhead.

SolastaCommunityExpansion/Classes/Witch.cs
RealBazou: Again, I have started using the API for this, but I find that sometimes it's easier for me to work from a copy that I want to edit surgically. The API sadly doesn't cover every attribute. I'm trying to use them as much as I can without creating overhead.

SolastaCommunityExpansion/Classes/Witch.cs
RealBazou: Thank you, this is now solved!

DubhHerder (commented):
You could try something like 


```using UnityEngine.AddressableAssets;

AssetReference assetReference = new AssetReference();
assetReference.SetField("m_AssetGUID", "763a4e34178a3224d9af53048f4a3e60");     spell.EffectDescription.EffectParticleParameters.SetTargetParticleReference(assetReference);
```
or maybe 
```
spell.EffectDescription.EffectParticleParameters.SetTargetParticleReference(new assetReference("763a4e34178a3224d9af53048f4a3e60"));
```
havent tested the code to confirm it works but it should be close.

SolastaCommunityExpansion/Classes/Witch.cs
RealBazou: Where is the method that does calculation for the enum? I couldn't find anything for this:
  public enum OriginOfAmount
  {
    None,
    SourceDamage,
    SourceGain,
    AddDice,
    Fixed,
    SourceHalfHitPoints,
    SourceSpellCastingAbility,
    SourceSpellAttack,
  }
Extending this could be really convenient

RealBazou: A questionable way of getting proficiency bonus would be to somehow do SourceSpellAttack - SourceSpellCastingAbility.
Also, getting the class/character level would be great.

DubhHerder: ApplySummonForm in RulesetImplementationManagerLocation.cs
If you're having trouble finding a particular piece of code, have you tried indexing code you've decompiled. 

For this particular case, yeah OriginOfAmount could be extended but the current options are reasonable approximations for 5xcharacter level and pb. I did a rough comparison when I was looking at it in the past

SolastaCommunityExpansion/Patches/ExtraSummoningModifiers/RulesetImplementationManagerLocation.cs
```diff
@@ -0,0 +1,172 @@
+using System.Collections.Generic;
```
RealBazou, line 1: I am unsure why, but spells that summon an item, like Goodberry or Produce Flame (Hold Flame) make the game crash. The prefix implementation is as follows:
- copied the entire code of the method
- added some cases in the switch case, using extended enum values
- used local vars for 
                var dummyCharacterList = new List<GameLocationCharacter>(); 
                var dummySizesList = new List<RulesetActor.SizeParameters>(); 
                var placementPositions = new List<int3>();
                var emptyFormationPositions = new List<int3>();
 since those were private members of the class, without a getter. I searched to see if this is used anywhere else in the class, but nothing came up.

SolastaCommunityExpansion/Patches/ExtraSummoningModifiers/RulesetImplementationManagerLocation.cs
```diff
@@ -0,0 +1,172 @@
...
+            {
+                if ((long) controlledEffectProxy.EffectGuid == (long) formsParams.activeEffect.Guid)
+                flag = true;
+            }
+            if (flag)
+                return false;
+            ServiceRepository.GetService<IGameLocationCharacterService>().CreateAndBindEffectProxy((RulesetActor) formsParams.sourceCharacter, formsParams.activeEffect, formsParams.position, element);
+            }
+            else
+            rulesetImplementationManager.ApplySummonForm(effectForm, formsParams);
```
RealBazou, line 163: I suspect the crash is here. How does one call the base class inside of a prefix method?

PhilPJL: You can't because ApplySummonForm is a virtual method.

Because RulesetImplementationManagerLocation.ApplySummonForm does this
image
If your parameters have the appropriate values in them you would end up calling base.ApplySummonForm.

Also, RulesetImplementationManager.ApplySummonForm only acts on it's parameters.  It may as well be a static method.  You can do this
`new RulesetImplementationManager().ApplySummonForm(effectForm, formsParams);`
if it's appropriate.

How can we repro your crash.  It should be obvious in the debugger.

RealBazou: I have attempted using a "ReversePatch" as documented here: https://harmony.pardeike.net/articles/patching-edgecases.html#calling-base-methods
It appears to work, however I am unsure if this is the correct approach. I am definitely not in my element here.

RealBazou: It was possible to reproduce the crash by using the Goodberry spell or Produce Flame -> Hold the Flame

SolastaCommunityExpansion/Helpers/Extensions.cs
```diff
@@ -20,4 +20,36 @@ internal static class Extensions
             return features.OfType<T>().ToList();
         }
     }
+
+    public enum ExtraRitualCasting
+    {
+        None = RuleDefinitions.RitualCasting.None,
+        Prepared = RuleDefinitions.RitualCasting.Prepared,
+        Spellbook = RuleDefinitions.RitualCasting.Spellbook,
+        Known = 1337
```
ChrisPJohn, line 29: Is there a reason leet was chosen for the number here and below? Is it just something high enough that the official TA enums are unlikely to reach that number?

SolastaCommunityExpansion/Models/PowersContext.cs
```diff
@@ -0,0 +1,77 @@
...
+                    Main.Settings.PowerEnabled.Add(powerName);
+                }
+            }
+            else
+            {
+                Main.Settings.PowerEnabled.Remove(powerName);
+            }
+        }
+
+        public static string GeneratePowersDescription()
```
ChrisPJohn, line 58: So these strings are designed to be used in the generation of the nexus mod description. You can add them similar to how the feats and subclasses are done here (this comment also applies to spells and classes):

https://github.com/ChrisPJohn/SolastaCommunityExpansion/blob/Dev/SolastaCommunityExpansion/Viewers/HelpAndCreditsViewer.cs#L68-L73

SolastaCommunityExpansion/Powers/BazouPowers.cs
```diff
@@ -0,0 +1,58 @@
...
+        public static readonly Guid BAZOU_POWERS_BASE_GUID = new Guid("99cee84d-6187-4d7f-a36e-1bd96d3f2deb");
+
+        public static void CreatePowers(List<FeatureDefinitionPower> powers)
+        {
+
+            powers.Add(BuildHelpAction());
+
+        }
+
+        private static FeatureDefinitionPower BuildHelpAction()
```
ChrisPJohn, line 20: Maybe I missed it somewhere, but should all characters be given this power?

ChrisPJohn (changes requested):
I'm still working my way through. There's a lot in here. I'll do more tomorrow. Here's a few comments though.

SolastaCommunityExpansion/Patches/ExtraSummoningModifiers/RulesetImplementationManagerLocation.cs
ChrisPJohn: So if I understand this patch you want to calculate sourceAmount differently than the existing enum options do. I'm not a fan of a prefix that entirely replaces the method because that could cause issues for other mods/patches as well as be difficult to maintain with future game patches.

There are two alternative approaches:
1) Use a transpiler (https://harmony.pardeike.net/articles/patching-transpiler.html) @ThyWoof is more familiar with these than me.
2) Use a prefix patch on ApplySummonForm to calculate the sourceAmount and then store that. Then also add a prefix patch on characterMonster.InflictCondition that uses the stored value instead of the parameter.

I'm probably not explaining 2 well here, but something similar to what was done here:
First patch store a value: https://github.com/ChrisPJohn/SolastaCommunityExpansion/blob/843b77a15025e298497720c3a12fce2707f88706/SolastaCommunityExpansion/Patches/GameUiCharactersPanel/CharacterEditionScreenPatcher.cs#L14-L21
Second patch, if value is set, use it: https://github.com/ChrisPJohn/SolastaCommunityExpansion/blob/0f588f62fa7aece847c9ede2b8dedc0aaea1829e/SolastaCommunityExpansion/Patches/GameUiCharactersPanel/CharactersPanelPatcher.cs#L24-L28

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -0,0 +1,1290 @@
...
+            effectDescription.SetDurationType(RuleDefinitions.DurationType.Round);
+            effectDescription.SetEndOfEffect(RuleDefinitions.TurnOccurenceType.EndOfTurn);
+            effectDescription.SetHasSavingThrow(false);
+            effectDescription.SetRangeType(RuleDefinitions.RangeType.Self);
+// Target by tag?
+//            effectDescription.SetTargetFilteringTag(RuleDefinitions.TargetFilteringTag.CursedByMalediction);
+            effectDescription.SetTargetType(RuleDefinitions.TargetType.Sphere);
+            effectDescription.SetTargetParameter(12);
+            effectDescription.EffectForms.Clear();
+// Can we add a Condition dynamically? i.e. can we detect what kind of condition a creature has, and then add that condition here?
```
ChrisPJohn, line 956: There is a feat that adds a condition conditionally:

https://github.com/ChrisPJohn/SolastaCommunityExpansion/blob/Dev/SolastaCommunityExpansion/Feats/ElAntoniousFeats.cs#L41-L72

That is a little bit different that what you're describing, and I'm unsure if that code is triggered on an attack/power that doesn't cause damage. It also doesn't know what attack/spell/power triggered the hit at that point. I have been contemplating a more general CustomEffectForm that could be used to make powers/spells more flexible, but we don't have an implementation yet (partially because we haven't come across something that truly needed it yet).

SolastaCommunityExpansion/Classes/AbstractClass.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace SolastaCommunityExpansion.Classes
+{
+    internal abstract class AbstractClass
+    {
+        internal abstract CharacterClassDefinition GetClass();
+        internal abstract void BuildEquipment(CharacterClassDefinitionBuilder classBuilder);
+        internal abstract void BuildProficiencies(CharacterClassDefinitionBuilder classBuilder);
+        internal abstract void BuildSubclasses(CharacterClassDefinitionBuilder classBuilder);
+        internal abstract void BuildProgression(CharacterClassDefinitionBuilder classBuilder);
+        internal abstract void BuildClassStats(CharacterClassDefinitionBuilder classBuilder);
```
ChrisPJohn, line 12: These seem like private methods. As far as I can tell there's no need for anything outside of the Class to need to call these. If that's the case they shouldn't be in the AbstractClass.

SolastaCommunityExpansion/Spells/BazouSpells.cs
```diff
@@ -0,0 +1,421 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace SolastaCommunityExpansion.Spells
+{
+    internal static class BazouSpells
+    {
+        public static readonly Guid BAZOU_SPELLS_BASE_GUID = new Guid("91384db5-6659-4384-bf2c-3a41160343f4");
+
+        public static void CreateSpells(List<SpellDefinition> spells)
```
ChrisPJohn, line 14: Should any of these spells be added to spell lists for other classes?

ChrisPJohn (changes requested):

SolastaCommunityExpansion/Classes/Witch.cs
ChrisPJohn: And I threw together a quick prototype of custom effect forms:

https://github.com/ChrisPJohn/SolastaCommunityExpansion/pull/293

I haven't tested it at all, but it should support what you're looking to do with adding a condition dynamically.

SolastaCommunityExpansion/Classes/AbstractClass.cs
RealBazou: No problem I can remove them. My reasoning was that every class will need to define these, so I included them do sd to be defined in each class. I'll define them as private.

SolastaCommunityExpansion/Helpers/Extensions.cs
RealBazou: It was to be high enough as to not impede on TA values, and having to pick a number, I chose an eye-rollingly dumb and ironic one. We can change this to something more appropriate.

SolastaCommunityExpansion/Powers/BazouPowers.cs
RealBazou: In theory, yes, but I didn't want to impede on anyone's code, and I wasn't sure if this is the correct approach. I added it for the Witch class, but in theory, this should be in the default set of actions a character can have. This, and the missing grapple action.

SolastaCommunityExpansion/Spells/BazouSpells.cs
RealBazou: I'll check the official list in the book, but I think some classes should be getting these spells as well, yes. I'm not sure how to add these though. A bit like the Help action above, I'm not sure what is the correct approach to do this well.

RealBazou: - Eldritch Orb: Warlock
- Find Familiar: Wizard
- Frenzy: Bard, Wizard
- Minor Lifesteal: Sorcerer, Wizard
- Petal Storm: Druid
- Protect Threshold: Sorcerer, Wizard

SolastaCommunityExpansion/Spells/BazouSpells.cs
```diff
@@ -0,0 +1,421 @@
...
+            spell.EffectDescription.SetRangeParameter(1);
+            spell.EffectDescription.SetDurationType(RuleDefinitions.DurationType.Minute);
+            spell.EffectDescription.SetDurationParameter(10);
+            spell.EffectDescription.SetTargetType(RuleDefinitions.TargetType.Sphere);
+            spell.EffectDescription.SetTargetParameter(0);
+            spell.EffectDescription.SetHasSavingThrow(true);
+            spell.EffectDescription.SetSavingThrowAbility(AttributeDefinitions.Wisdom);
+            spell.EffectDescription.SetRecurrentEffect(RuleDefinitions.RecurrentEffect.OnEnter);
+
+// How can I get an Asset Reference if it is a GameObject loaded in the DB?
```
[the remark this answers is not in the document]
DubhHerder: Moved this comment because it was misplaced when initially added. 

You could also use the asset reference guids from the blueprints


```using UnityEngine.AddressableAssets;

AssetReference assetReference = new AssetReference();
assetReference.SetField("m_AssetGUID", "763a4e34178a3224d9af53048f4a3e60");     spell.EffectDescription.EffectParticleParameters.SetTargetParticleReference(assetReference);
```
or 
```
spell.EffectDescription.EffectParticleParameters.SetTargetParticleReference(new assetReference("763a4e34178a3224d9af53048f4a3e60"));
```
The previous solution is probably better, this is included for completeness

SolastaCommunityExpansion/Subclasses/Witch/WhiteWitch.cs
```diff
@@ -0,0 +1,110 @@
...
+    internal class WhiteWitch : AbstractSubclass
+    {
+        private CharacterSubclassDefinition Subclass;
+        internal override FeatureDefinitionSubclassChoice GetSubclassChoiceList()
+        {
+            DatabaseRepository.GetDatabase<FeatureDefinitionSubclassChoice>().TryGetElement("SubclassChoiceWitchCovens", out FeatureDefinitionSubclassChoice featureDefinitionSubclassChoice);
+            return featureDefinitionSubclassChoice;
+        }
+
+        internal override CharacterSubclassDefinition GetSubclass()
```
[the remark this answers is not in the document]
RealBazou: Removed in commit 6d543dc64afc57d98519eb34d93f765d210c795b: https://github.com/ChrisPJohn/SolastaCommunityExpansion/pull/206/commits/6d543dc64afc57d98519eb34d93f765d210c795b

SolastaCommunityExpansion/Spells/BazouSpells.cs
ThyWoof: I can help here. Will start thinking on an UI for that

SolastaCommunityExpansion/Spells/BazouSpells.cs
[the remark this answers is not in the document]
RealBazou: I have tried using the following:
`            // While the Game Object is set correctly, the particle effect doesn't show. 
            spell.EffectDescription.EffectParticleParameters.SetTargetParticleReference(DatabaseHelper.SpellDefinitions.ConjureAnimalsOneBeast.EffectDescription.EffectParticleParameters.GetField<UnityEngine.AddressableAssets.AssetReference>("targetParticleReference"));
`
The object was correctly set, however I didn't get the effect I was hoping for, i.e. the "summoning" particle on the floor. I tried changing the target type to "position", and then I do get that particle, but then I wouldn't get the Spike Growth particle itself. I finally decided to leave it as is, i.e. Sphere target type, with a size of 0, i.e. a weird "dot" that targets a grid square.

SolastaCommunityExpansion/Subclasses/Witch/WhiteWitch.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            GuiPresentation blank = new GuiPresentationBuilder("Feature/&NoContentTitle", "Feature/&NoContentTitle").Build();
+
+            var preparedSpells = new FeatureDefinitionAutoPreparedSpellsBuilder(
+                    "WhiteMagicAutoPreparedSpell",
+                    GuidHelper.Create(WW_BASE_GUID, "WhiteMagicAutoPreparedSpell").ToString(),
+                    new List<FeatureDefinitionAutoPreparedSpells.AutoPreparedSpellsGroup>{
+                            FeatureDefinitionAutoPreparedSpellsBuilder.BuildAutoPreparedSpellGroup(
+                                    1,
+                                    new List<SpellDefinition>{
```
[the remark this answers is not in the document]
RealBazou: I have fixed this and followed the standard for multiclass.

SolastaCommunityExpansion/Models/PowersContext.cs
RealBazou: I have added the strings in the HelpAndCreditsViewer file, following the same pattern as Feats.

SolastaCommunityExpansion/Spells/BazouSpells.cs
ChrisPJohn: Can we add either a todo here or an issue to the github repository about making these spells available to other classes?

SolastaCommunityExpansion/Classes/Witch.cs
```diff
@@ -0,0 +1,1290 @@
...
+            var classSpellCast = new CastSpellBuilder(
+                    "WitchSpellCast", 
+                    GuidHelper.Create(WITCH_BASE_GUID, "WitchSpellCast").ToString());
+
+            classSpellCast.SetGuiPresentation(
+                    new GuiPresentationBuilder(
+                            "Class/&WitchSpellcastingDescription",
+                            "Class/&WitchSpellcastingTitle")
+                            .Build());
+            classSpellCast.SetKnownCantrips(new List<int>{
```
ChrisPJohn, line 407: I believe you could use some helper functions to simplify this section:

```
classSpellCast.SetKnownCantrips(4, 1, CasterProgression.FULL_CASTER);
classSpellCast.SetKnownSpells(2, 1,  CasterProgression.FULL_CASTER);
classSpellCast.SetSlotsPerLevel(1,  CasterProgression.FULL_CASTER);

```

This is optional to change, but since it looks like your following the standard spell progression I just wanted to call out the helper functions that help make this simpler for you.

SolastaCommunityExpansion/Powers/BazouPowers.cs
ChrisPJohn: Can we add either a todo here to add this to characters or an issue in the repository?

ChrisPJohn (changes requested):
Several small things like adding todos or issues to document future work. My main concern though is the patch in RulesetImplementationManagerLocation.cs. I'd like to see something less likely to cause maintenance issues in the future and conflicts with other mods.

I do wonder if we can get the desired effect with a prefix on ApplySummonForm to calculate the sourceAmount for the new cases and then a prefix on InflictCondition that replaces the sourceAmount with our precalculated amount.

ChrisPJohn (approved):

PR #208, opened 2021-12-12
Save by location changes

PhilPJL:
Changed save by location to
1) save locations in their own folder (used to save in the first folder of a linked set)
2) save locations and campaigns in the following folder structure

Solasta\Saves\Ce\Location\<loc names>
Solasta\Saves\Ce\Campaign\<campaign names>
Solasta\Saves ... main campaign saves

Drop down is now colour coded to distinguish campaigns from locations.  Could use some suitable icons.
image


ThyWoof:
should we add support for the upcoming campaign? keep all official saves as is but would need to change that constant to an array with official campaign names...

ThyWoof (approved):

PR #225, opened 2021-12-13
finishes the controllers model migration

ThyWoof:


SolastaCommunityExpansion/Models/PlayerControllerContext.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        private static bool PlayerInControlOfEnemy { get; set; }
+
+        internal static readonly string[] Controllers = new string[] { "Human", "Computer" };
+
+        internal static readonly int[] ControllersChoices = new int[Settings.MAX_PARTY_SIZE];
+
+        internal static bool IsOffGame => Gui.Game == null;
+
+        internal static bool IsMultiplayer => ServiceRepository.GetService<INetworkingService>().IsMultiplayerGame;
```
PhilPJL, line 21: Is it safe to call ServiceRepository.GetService from a static method/property?

PhilPJL: e.g. one the game is started could the network service change from null to not null or vice versa?

PhilPJL: Never mind - it's always available

SolastaCommunityExpansion/Patches/PlayerController/GameLocationCharacterManagerPatcher.cs
```diff
@@ -1,18 +1,18 @@
 ﻿using System.Diagnostics.CodeAnalysis;
 using HarmonyLib;
 
-namespace SolastaCommunityExpansion.Patches.HeroController
+namespace SolastaCommunityExpansion.Patches.PlayerController
 {
     //
-    // TODO: double-check if this indeed required on latest game version. The test scenario is a hero dies when under AI control.
+    // TODO: double-check if this indeed required on latest game version. The test scenario is a hero dies when under AI control or an enemy dies when under player control
```
PhilPJL, line 7: Are you double checking?

PhilPJL (changes requested):
need to complete your TODO: or remove

SolastaCommunityExpansion/Patches/PlayerController/GameLocationCharacterManagerPatcher.cs
ThyWoof: I'll once I have a chance but as far as I can remember this patch should be there...

ThyWoof: when I set logs the sequence is:

1. turn_start
2. optional kill_character
3. turn_end or turn_victory

these patches are here to support these scenarios:

1. Gamers doing a PvP Arena battle. This is when you can select Heroes from Encounters > Characters Pool and Spawn them in an encounter. When I wrote these game had a logic to game over if a hero dies. any hero.
2. What I'm not sure here is if I also need to call stop on heroes controlled by computer.

ThyWoof: 2) isn't a problem as both Stop implementations are safe to call multiple times in sequence. when that happens state stays the same.

PhilPJL (approved):
Leaving TODO for now

PR #229, opened 2021-12-13
messing about

PhilPJL:
I was experimenting...  Not for merging.

I thought an interface ISubclass would be better than an abstract class.  For example MasterManipulator.

However I then wondered why these classes aren't static like everything else since presumably we only want one instance.  See SpellMaster.

Is there a way to invite conversations on code without creating a PR?


ChrisPJohn:
> I was experimenting... Not for merging.
> 
> I thought an interface ISubclass would be better than an abstract class. For example MasterManipulator.
> 
> However I then wondered why these classes aren't static like everything else since presumably we only want one instance. See SpellMaster.
> 
> Is there a way to invite conversations on code without creating a PR?

When I was first adding the subclasses I had the show/hide logic in the abstract subclass class. Since that was all moved out, it can be an interface.

As for static vs not, I don't have a strong preference. Since I was extending the abstract class initially they couldn't be static.

Issue #78, opened 2021-11-13
Level 20 features: Barbarian: Indomitable Might


ChrisPJohn:

DubhHerder:
**Feature :**
Beginning at 18th level, if your total for a Strength check is less than your Strength score, you can use that score in place of the total.

RedOrcaCode:
I'm going to take a look at this one.

PR #238, opened 2021-12-16
Indomitable Might barbarian class feature

RedOrcaCode:
Adds the 18th level barbarian class feature Indomitable Might (#78). Tested with shove, escaping from _black tentacles_, and interacting with roll activators in custom dungeons.

### DIFF
--- a/SolastaCommunityExpansion/Level20/Features/IndomitableMightBuilder.cs
+++ b/SolastaCommunityExpansion/Level20/Features/IndomitableMightBuilder.cs
@@ -29,12 +29,7 @@ namespace SolastaCommunityExpansion.Level20.Features
     {
         public int? MinimumStrengthAbilityCheckTotal(RulesetCharacter character, string proficiencyName)
         {
-            if (character == null)
-            {
-                return null;
-            }
-
-            return character.GetAttribute(AttributeDefinitions.Strength).CurrentValue;
+            return character?.GetAttribute(AttributeDefinitions.Strength).CurrentValue;
         }
     }
 }
--- a/SolastaCommunityExpansion/Level20/Features/IndomitableMightBuilder.cs
+++ b/SolastaCommunityExpansion/Level20/Features/IndomitableMightBuilder.cs
@@ -1,0 +1,40 @@
+﻿using SolastaCommunityExpansion.CustomFeatureDefinitions;
+using SolastaModApi;
+using SolastaModApi.Extensions;
+
+namespace SolastaCommunityExpansion.Level20.Features
+{
+    internal class IndomitableMightBuilder : BaseDefinitionBuilder<IndomitableMight>
+    {
+        private const string IndomitableMightName = "BarbarianIndomitableMight";
+        private const string IndomitableMightGuid = "2a0e9082-c81d-4d02-800a-92f04fbe85dc";
+
+        protected IndomitableMightBuilder(string name, string guid) : base(name, guid)
+        {
+            var guiPresentationBuilder = new GuiPresentationBuilder(
+                "Feature/&BarbarianIndomitableMightDescription",
+                "Feature/&BarbarianIndomitableMightTitle");
+
+            Definition.SetGuiPresentation(guiPresentationBuilder.Build());
+        }
+
+        private static FeatureDefinition CreateAndAddToDB(string name, string guid)
+            => new IndomitableMightBuilder(name, guid).AddToDB();
+
+        internal static readonly FeatureDefinition IndomitableMight =
+            CreateAndAddToDB(IndomitableMightName, IndomitableMightGuid);
+    }
+    
+    internal class IndomitableMight : FeatureDefinition, IMinimumAbilityCheckTotal
+    {
+        public int? MinimumStrengthAbilityCheckTotal(RulesetCharacter character, string proficiencyName)
+        {
+            if (character == null)
+            {
+                return null;
+            }
+
+            return character.GetAttribute("Strength").CurrentValue;
+        }
+    }
+}
--- a/SolastaCommunityExpansion/Level20/Features/IndomitableMightBuilder.cs
+++ b/SolastaCommunityExpansion/Level20/Features/IndomitableMightBuilder.cs
@@ -6,7 +6,7 @@ namespace SolastaCommunityExpansion.Level20.Features
 {
     internal class IndomitableMightBuilder : BaseDefinitionBuilder<IndomitableMight>
     {
-        private const string IndomitableMightName = "BarbarianIndomitableMight";
+        private const string IndomitableMightName = "ZSBarbarianIndomitableMight";
         private const string IndomitableMightGuid = "2a0e9082-c81d-4d02-800a-92f04fbe85dc";
 
         protected IndomitableMightBuilder(string name, string guid) : base(name, guid)
@@ -34,7 +34,7 @@ namespace SolastaCommunityExpansion.Level20.Features
                 return null;
             }
 
-            return character.GetAttribute("Strength").CurrentValue;
+            return character.GetAttribute(AttributeDefinitions.Strength).CurrentValue;
         }
     }
 }
--- a/SolastaCommunityExpansion/Patches/Level20/RulesetActorPatcher.cs
+++ b/SolastaCommunityExpansion/Patches/Level20/RulesetActorPatcher.cs
@@ -1,0 +1,54 @@
+﻿using HarmonyLib;
+using System.Collections.Concurrent;
+using System.Diagnostics.CodeAnalysis;
+
+namespace SolastaCommunityExpansion.Patches.Level20
+{
+    [HarmonyPatch(typeof(RulesetActor), "RollDie")]
+    [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
+    internal static class RulesetActor_RollDie
+    {
+        private static readonly ConcurrentDictionary<RulesetActor, int> NextAbilityCheckMinimum = new ConcurrentDictionary<RulesetActor, int>();
+
+        internal static void Postfix(
+            RulesetActor __instance,
+            RuleDefinitions.DieType dieType,
+            RuleDefinitions.RollContext rollContext,
+            ref int __result)
+        {
+            if (!Main.Settings.EnableLevel20)
+            {
+                return;
+            }
+
+            if (dieType != RuleDefinitions.DieType.D20 || rollContext != RuleDefinitions.RollContext.AbilityCheck)
+            {
+                return;
+            }
+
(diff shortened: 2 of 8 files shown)


SolastaCommunityExpansion/Patches/Level20/RulesetCharacterPatcher.cs
```diff
@@ -0,0 +1,143 @@
...
+                string opponentProficiencyName)
+            {
+                if (!Main.Settings.EnableLevel20)
+                {
+                    return;
+                }
+
+                // ResolveContestCheck calls RulesetActor.RollDie twice (once for you, once for your opponent).
+                // SetNextAbilityCheckMinimum below tells the RulesetActor.RollDie patch that the next call to RollDie (for the specified RulesetCharacter) must have a certain minimum value.
+
```
RedOrcaCode, line 65: As the above comment shows, this is where it gets complicated. Shoves are handled via RulesetCharacter.ResolveContestCheck, which in turn calls RulesetActor.RollDie twice (once for you, once for your opponent).

Neither method accepts a minimum roll value argument, so the only option is to force RollDie to return the die value we want. Unfortunately, RollDie doesn't know the roll modifiers - it only handles the d20 itself. In order to get knowledge of the roll modifiers into the RollDie patch, I use RulesetActor_RollDie.SetNextAbilityCheckMinimum. That forces the next call of RollDie (by a certain character) to return the value we want.

SolastaCommunityExpansion/Patches/Level20/RulesetCharacterPatcher.cs
```diff
@@ -0,0 +1,143 @@
...
+                }
+
+                int? modifiedMinRoll = MinimumStrengthAbilityCheckDieRoll(__instance, baseBonus, rollModifier, proficiencyName);
+
+                if (modifiedMinRoll.HasValue && modifiedMinRoll.Value > minRoll)
+                {
+                    minRoll = modifiedMinRoll.Value;
+                }
+            }
+        }
```
RedOrcaCode, line 40: It should be easy to adapt this and make minimum totals work for most ability checks, but I only tested it on Strength. Unfortunately, not all checks use RollAbilityCheck. Shoves use ResolveContestCheck instead. I wasn't confident that I had covered every scenario (besides Strength), so I made this Strength-only for now.

SolastaCommunityExpansion/Patches/Level20/RulesetCharacterPatcher.cs
```diff
@@ -0,0 +1,143 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace SolastaCommunityExpansion.Patches.Level20
+{
+    internal static class RulesetCharacterPatcher
+    {
+        [HarmonyPatch(typeof(RulesetCharacter), "RollAbilityCheck")]
+        [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
+        internal static class RulesetCharacter_RollAbilityCheck
```
RedOrcaCode, line 13: It should be easy to adapt this and make minimum totals work for most ability checks, but I only tested it on Strength. Unfortunately, not all checks use RollAbilityCheck. Shoves use ResolveContestCheck instead. I wasn't confident that I had covered every scenario (besides Strength), so I made this Strength-only for now.

SolastaCommunityExpansion/Level20/Features/IndomitableMightBuilder.cs
```diff
@@ -0,0 +1,40 @@
+﻿using SolastaCommunityExpansion.CustomFeatureDefinitions;
+using SolastaModApi;
+using SolastaModApi.Extensions;
+
+namespace SolastaCommunityExpansion.Level20.Features
+{
+    internal class IndomitableMightBuilder : BaseDefinitionBuilder<IndomitableMight>
+    {
+        private const string IndomitableMightName = "BarbarianIndomitableMight";
```
ThyWoof, line 9: Please Prefix the name with ZS. This will avoid any future conflicts when / if TA decides to implement it.

SolastaCommunityExpansion/Patches/Level20/RulesetActorPatcher.cs
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+
+        internal static void SetNextAbilityCheckMinimum(RulesetActor actor, int minimum)
+        {
+            if (actor == null)
+            {
+                return;
+            }
+
```
ThyWoof, line 50: What would happen here with a Multiclass Druid 2 / Barb 18 while in wildshape? Is it possible for this tab to get out of sync as game substitutes the hero with a monster?

ThyWoof (commented):
Also don’t forget to edit the help and credits display to remove the planned tag at this feature.

SolastaCommunityExpansion/Patches/Level20/RulesetCharacterPatcher.cs
```diff
@@ -0,0 +1,143 @@
...
+                {
+                    int? instanceMinRoll = MinimumStrengthAbilityCheckDieRoll(__instance, baseBonus, rollModifier, proficiencyName);
+
+                    if (instanceMinRoll.HasValue)
+                    {
+                        RulesetActor_RollDie.SetNextAbilityCheckMinimum(__instance, instanceMinRoll.Value);
+                    }
+                }
+
+                if (opponentAbilityScoreName == "Strength")
```
PhilPJL, line 76: Ideally we should use `AttributeDefinitions.Strength` not "Strength" etc..

SolastaCommunityExpansion/Patches/Level20/RulesetCharacterPatcher.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        private static int? MinimumStrengthAbilityCheckTotal(RulesetCharacter character, string proficiencyName)
+        {
+            var featuresToBrowse = new List<FeatureDefinition>();
+
+            character.EnumerateFeaturesToBrowse<IMinimumAbilityCheckTotal>(featuresToBrowse);
+
+            int? minimumTotal = null;
+
+            foreach (FeatureDefinition feature in featuresToBrowse)
```
PhilPJL, line 115: You can use Linq to good effect here.  Something like

`return featuresToBrowse.OfType<IMinimumAbilityCheckTotal>().Select(f => (int?)f.MinimumStrengthAbilityCheckTotal(character, proficiencyName)).Max();`

SolastaCommunityExpansion/Level20/Features/IndomitableMightBuilder.cs
RedOrcaCode: Done.

SolastaCommunityExpansion/Patches/Level20/RulesetCharacterPatcher.cs
RedOrcaCode: Oh nice - I didn't realize those existed.

SolastaCommunityExpansion/Patches/Level20/RulesetCharacterPatcher.cs
RedOrcaCode: That's a lot cleaner, thanks. Updated.

SolastaCommunityExpansion/Patches/Level20/RulesetActorPatcher.cs
RedOrcaCode: It shouldn't be possible for them to get out of sync, since there's no way to wild shape between the beginning of ResolveContestCheck and the time that it calls RollDie. ResolveContestCheck works something like this:

```
public int ResolveContestCheck( /* SNIP */ )
{
    /* SNIP - checking advantage/proficiency; no return statements */
    int rawRoll = this.RollDie( /* SNIP */ );
    /* SNIP - checking advantage/proficiency for opponent; no return statements */
    int opponentRawRoll = opponent.RollDie( /* SNIP */ )
   /* SNIP - computing and returning the result */
}
```

The only way for RollDie not to get called is if ResolveContestCheck throws an exception, which would probably cause the game to crash. Otherwise, RollDie has to be called before you have any chance to wild shape.

SolastaCommunityExpansion/Level20/Features/IndomitableMightBuilder.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        internal static readonly FeatureDefinition IndomitableMight =
+            CreateAndAddToDB(IndomitableMightName, IndomitableMightGuid);
+    }
+    
+    internal class IndomitableMight : FeatureDefinition, IMinimumAbilityCheckTotal
+    {
+        public int? MinimumStrengthAbilityCheckTotal(RulesetCharacter character, string proficiencyName)
+        {
+            if (character == null)
```
PhilPJL, line 32: You could implement this whole method as

return character?.GetAttribute(AttributeDefinitions.Strength).CurrentValue;

Why is there the parameter 'proficiencyName'?

PhilPJL (commented):
I'll think Chris is the best person to approve this.

SolastaCommunityExpansion/Patches/Level20/RulesetActorPatcher.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Collections.Concurrent;
+using System.Diagnostics.CodeAnalysis;
+
+namespace SolastaCommunityExpansion.Patches.Level20
+{
+    [HarmonyPatch(typeof(RulesetActor), "RollDie")]
+    [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
+    internal static class RulesetActor_RollDie
+    {
+        private static readonly ConcurrentDictionary<RulesetActor, int> NextAbilityCheckMinimum = new ConcurrentDictionary<RulesetActor, int>();
```
PhilPJL, line 11: Is there any multi-threading in the game?

RedOrcaCode: I don't think so - at least not for game logic - but I wasn't sure so I figured it wouldn't hurt.

SolastaCommunityExpansion/Level20/Features/IndomitableMightBuilder.cs
RedOrcaCode: I included proficiencyName to allow for more precise control over similar features in the future. If you wanted to create a feature like Indomitable Might that only worked on Athletics checks, it supports that.

ThyWoof (commented):
I’m ok with the changes but prefer @ChrisPJohn to give his final thought

ChrisPJohn (approved):

PR #252, opened 2021-12-18
keeps char. panel open w/ hero selected after level up / char creation

ThyWoof:
. quality of life patch to keep the char panel open on main screen after char. creation / level up.
. the first patch is protected by an invisible toggle on first state
. the 2 other patches are protected by a property set on first patch


ThyWoof:
I believe this provides a good balance as although not difficult to add a Level Up button on the UI it proven very difficult to trigger a new level up with all the async events happening behind the scene during the screen unload.

ChrisPJohn (approved):

PhilPJL:
Agree that this is good enough.  One extra tweak would be good though.  Ensure the selected character is visible in the list after level up.  Created a character 'zzz' which remained selected but off the screen.

PR #272, opened 2021-12-20
add more sorting / tweak existing ones

ThyWoof:


ThyWoof:
using a Postfix on Compare() will make the game calls CompareTo twice. On original sort and on patched one. I still prefer a Postfix vs. a bool Prefix to keep compatibility with any other mod out there.

PhilPJL (approved):

PR #283, opened 2021-12-22
patches protection review

ThyWoof:
Reviewed all patches
Added comments to the ones that don't need protection
Consolidated BugFix patches under same namespace / folder
Better names on hidden settings
No significant code changes except for if / return when necessary

### DIFF
--- a/SolastaCommunityExpansion/Models/DungeonMakerContext.cs
+++ b/SolastaCommunityExpansion/Models/DungeonMakerContext.cs
@@ -15,7 +15,7 @@ namespace SolastaCommunityExpansion.Models
             UpdatePropsPlacement();
         }
 
-        internal static void BackupAndDelete(string path, UserContent userContent)
+        public static void BackupAndDelete(string path, UserContent userContent)
         {
             var backupDirectory = Path.Combine(Main.MOD_FOLDER, BACKUP_FOLDER);
 
--- a/SolastaCommunityExpansion/Patches/Bugfix/GameLoreManagerPatcher.cs
+++ b/SolastaCommunityExpansion/Patches/Bugfix/GameLoreManagerPatcher.cs
@@ -1,8 +1,11 @@
 ﻿using System.Diagnostics.CodeAnalysis;
 using HarmonyLib;
 
-namespace SolastaCommunityExpansion.Patches.LoadFixers
+namespace SolastaCommunityExpansion.Patches.BugFix
 {
+    //
+    // TODO @IMPPHIL: should we protect
+    //
     [HarmonyPatch(typeof(GameLoreManager), "SerializeElements")]
     [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
     internal static class GameLoreManager_SerializeElements
@@ -13,6 +16,11 @@ namespace SolastaCommunityExpansion.Patches.LoadFixers
         // from the list of known recipes.
         internal static void Postfix(GameLoreManager __instance)
         {
+            if (!Main.Settings.BugFixNullRecipesOnGameSerialization)
+            {
+                return;
+            }
+
             __instance.KnownRecipes.RemoveAll(item => item == null);
         }
     }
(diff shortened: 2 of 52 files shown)


SolastaCommunityExpansion/Models/DungeonMakerContext.cs
```diff
@@ -15,7 +15,7 @@ internal static void Load()
             UpdatePropsPlacement();
         }
 
-        internal static void BackupAndDelete(string path, UserContent userContent)
+        public static void BackupAndDelete(string path, UserContent userContent)
```
ThyWoof, line 18: made this public to remove the Bindings and #pragma on the Transpiler side.

SolastaCommunityExpansion/Patches/LongActivationPowers/GameLocationCharacterPatcher.cs
```diff
@@ -3,7 +3,10 @@
 
 namespace SolastaCommunityExpansion.Patches.LongActivationPowers
 {
-    // Yes the actual game typos this it is "OnPower" and not the expected "OnePower".
+    // Yes the actual game typos this it is "OnPower" and not the expected "OnePower"
+    //
+    // this patch shouldn't be protected
```
ChrisPJohn, line 8: If we want to protect as much as possible, this could be a hidden setting default to true.

Turning this off means that characters who only have powers with an activation time longer than a minute can't activate the powers. Effectively it makes several feats almost entirely useless.

SolastaCommunityExpansion/Patches/PowerSharedPool/RulesetCharacterPatcher.cs
```diff
@@ -9,6 +9,9 @@ namespace SolastaCommunityExpansion.Patches.PowerSharedPool
 {
     internal static class RulesetCharacterPatcher
     {
+        //
+        // this patch shouldn't be protected
```
ChrisPJohn, line 13: The patches here are effectively protected by 
```
if (!(modifiedPower.PowerDefinition is IPowerSharedPool))
```

in the UpdateUsageForPowerPool function

SolastaCommunityExpansion/Patches/ClassHoldingFeature/RulesetCharacterHeroPatcher.cs
```diff
@@ -9,7 +9,7 @@ namespace SolastaCommunityExpansion.Patches.ClassHoldingFeature
     internal static class RulesetCharacterHero_FindClassHoldingFeature
     {
         //
-        // TODO @CHRIS: WE SHOULD PROTECT THIS PATCH
+        // TODO @CHRIS: should we protect this patch? What is the purpose?
```
ChrisPJohn, line 12: It was added to support the Path of Light Barbarian:

https://github.com/ChrisPJohn/SolastaCommunityExpansion/blob/Dev/SolastaCommunityExpansion/Subclasses/Barbarian/PathOfTheLight.cs#L531

My understanding is it's a way to make sure the barbarian feature only scales with barbarian levels and not character levels. I don't know if multi class mod handles this already, if so we can probably get rid of it.

SolastaCommunityExpansion/Patches/Bugfix/GameLoreManagerPatcher.cs
```diff
@@ -13,6 +16,11 @@ internal static class GameLoreManager_SerializeElements
         // from the list of known recipes.
         internal static void Postfix(GameLoreManager __instance)
         {
+            if (!Main.Settings.BugFixNullRecipesOnGameSerialization)
```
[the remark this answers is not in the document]
PhilPJL: Agree with Chris

SolastaCommunityExpansion/Patches/Camera/CameraControllerLocationPatcher.cs
```diff
@@ -16,41 +16,38 @@ internal static class CameraControllerLocation_FollowCharacterForBattle
         /// </summary>
         internal static bool Prefix(CameraControllerLocation __instance, GameLocationCharacter character)
         {
-            //
-            // TODO @IMPPHIL: FIX EXIT ROUTE
```
PhilPJL, line 20: ok will do

PhilPJL (approved):

PR #288, opened 2021-12-23
Camp and entrance/exit map icons

### LINKED ISSUE #139: Additional icons on the level map
We get icons for discovered chests like this:
image

Add icons for discovered entrance, exit, camp fires.

PhilPJL:
Closes #139


ThyWoof (approved):

ThyWoof:
Approved but wondering if this setting should be under Dungeon Maker? Does it also work in regular campaigns?

PhilPJL:
> Approved but wondering if this setting should be under Dungeon Maker? Does it also work in regular campaigns?

Good question.  I had only tested this on custom dungeons, e.g. with Ruins of Ilthismar
image

Looks like only entrance/exit show up in main campaign because long rests are already handled
image

We can refine this if necessary.

PR #291, opened 2021-12-23
fix sorting by hero level to work on both SC / MC builds

ThyWoof:


SolastaCommunityExpansion/Patches/GameUiCharactersPanel/CharacterFilteringGroupPatcher.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Reflection.Emit;
+using HarmonyLib;
+
+namespace SolastaCommunityExpansion.Patches.GameUiCharactersPanel
+{
+    internal static class CharacterFilteringGroupPatcher
```
PhilPJL, line 9: You could do this with a postfix.  Is that better or worse?

if(sortCategory == CharacterLevel) {...}

SolastaCommunityExpansion/Patches/GameUiCharactersPanel/CharacterSelectionModalPatcher.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Reflection.Emit;
+using HarmonyLib;
+
+namespace SolastaCommunityExpansion.Patches.GameUiCharactersPanel
+{
+    internal static class CharacterSelectionModalPatcher
```
PhilPJL, line 9: You could do this with a prefix/postfix which cached the heroes levels, set level[0] = sum and then restored in the postfix.  Is that better or worse than a transpiler?

SolastaCommunityExpansion/Patches/GameUiCharactersPanel/CharacterFilteringGroupPatcher.cs
ThyWoof: changed to postfix.

SolastaCommunityExpansion/Patches/GameUiCharactersPanel/CharacterSelectionModalPatcher.cs
ThyWoof: Worse as we now need 2 injection points. This is a very simple transpiler to fix in case it needs to be. I also don't foresee this code changing at all in upcoming DLCs.

ThyWoof:
@PhilPJL , consolidated one of the patches under the BugFix you had written. Left the original file with a postfix implementation in case we deprecate that bugfix one.

PhilPJL (approved):

PR #293, opened 2021-12-24
Rough idea for how to provide more flexibility in spell/power effects.

ChrisPJohn:
We probably shouldn't add this until we have a concrete use for this. Though from comments elsewhere I think there are potential uses.


SolastaCommunityExpansion/CustomFeatureDefinitions/CustomEffectForm.cs
```diff
@@ -0,0 +1,33 @@
...
+    class CustomEffectForm : EffectForm
+    {
+        public delegate bool ApplyDelegate(EffectForm effectForm,
+            RulesetImplementationDefinitions.ApplyFormsParams formsParams,
+            bool retargeting,
+            bool proxyOnly,
+            bool forceSelfConditionOnly);
+        private ApplyDelegate Apply;
+
+        internal void SetApplyDelegate(ApplyDelegate del)
```
PhilPJL, line 16: I would use `Func<EffectForm, Apply FormsParams, bool, bool, bool,, bool>` rather than a delegate

SolastaCommunityExpansion/Patches/CustomEffectForm/RulesetImplementationManagerPatcher.cs
```diff
@@ -0,0 +1,26 @@
...
+    [HarmonyPatch(typeof(RulesetImplementationManager), "ApplyEffectForms")]
+    internal static class RulesetImplementationManagerApplyCustomEffectForm
+    {
+        public static void Postfix(List<EffectForm> effectForms,
+            RulesetImplementationDefinitions.ApplyFormsParams formsParams,
+            bool retargeting = false,
+            bool proxyOnly = false,
+            bool forceSelfConditionOnly = false)
+        {
+           foreach(EffectForm effectForm in effectForms)
```
PhilPJL, line 17: `.OfType<>` ?

SolastaCommunityExpansion/Patches/CustomEffectForm/RulesetImplementationManagerPatcher.cs
```diff
@@ -0,0 +1,26 @@
...
+            {
+                if (effectForm is CustomFeatureDefinitions.CustomEffectForm customEffect)
+                {
+                    customEffect.ApplyForm(effectForm, formsParams, retargeting, proxyOnly, forceSelfConditionOnly);
```
PhilPJL, line 21: Is `effectForm `always going to be the same instance as `customEffect`? If so you don't need the `effectForm `parameter.

SolastaCommunityExpansion/CustomFeatureDefinitions/CustomEffectForm.cs
```diff
@@ -0,0 +1,33 @@
...
+            Apply = del;
+        }
+
+        public void ApplyForm(EffectForm effectForm,
+            RulesetImplementationDefinitions.ApplyFormsParams formsParams,
+            bool retargeting,
+            bool proxyOnly,
+            bool forceSelfConditionOnly)
+        {
+            if (Apply != null)
```
PhilPJL, line 27: Since your delegate returns a bool, do you want to return it here?

ChrisPJohn: The bool was a mistake, no need to return it.

SolastaCommunityExpansion/CustomFeatureDefinitions/CustomEffectForm.cs
ChrisPJohn: The bool return was a mistake, Func needs a return value, so I'm sticking with delegate.

ChrisPJohn: I also prefer to have the type declaration that a delegate provides because this is part of the interface of the custom effect form.

PhilPJL: You would use an Action<> if you don't need a return value.

PhilPJL: but delegate is fine

SolastaCommunityExpansion/CustomFeatureDefinitions/CustomEffectForm.cs
```diff
@@ -0,0 +1,33 @@
...
+    class CustomEffectForm : EffectForm
+    {
+        public delegate void ApplyDelegate(CustomEffectForm effectForm,
+            RulesetImplementationDefinitions.ApplyFormsParams formsParams,
+            bool retargeting,
+            bool proxyOnly,
+            bool forceSelfConditionOnly);
+        private ApplyDelegate Apply;
+
+        internal void SetApplyDelegate(ApplyDelegate del)
```
PhilPJL, line 16: Where do you intend `internal void SetApplyDelegate(ApplyDelegate del)` to be called from? Should it be protected?

Is CustomEffectForm a base class for CustomEffectForms?  If so make it abstract and give it a protected constructor which takes the delegate?

SolastaCommunityExpansion/Patches/CustomEffectForm/RulesetImplementationManagerPatcher.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using HarmonyLib;
+
+namespace SolastaCommunityExpansion.Patches.CustomEffectForm
+{
+    [HarmonyPatch(typeof(RulesetImplementationManager), "ApplyEffectForms")]
+    internal static class RulesetImplementationManagerApplyCustomEffectForm
+    {
+        public static void Postfix(List<EffectForm> effectForms,
```
PhilPJL, line 12: Ideally use IEnumerable<EffectForm> so clients can pass in whatever collection they like.

SolastaCommunityExpansion/CustomFeatureDefinitions/CustomEffectForm.cs
ChrisPJohn: SetApplyDelegate would be called from wherever the spell/power is getting defined that contains the custom effect. I wasn't intending to merge this until we had a solid use case of it (also currently this code is untested since I haven't set up a use for this yet).

I had intended it as you instantiate CustomEffectForm and set the apply delegate. I wasn't intending to make users define a class for each CustomEffectForm they declare (and if I was having them declare the class there's no need for a delegate). We could switch to this being an abstract class, I don't have a strong preference on that.

PhilPJL (dismissed):

ChrisPJohn:
I can't see this comment in the GitHub UI, it just came to my email. That
parameter is part of the TA defined function I believe I need to use the
exact same type as them for the patch to function so iEnumerable is not an
option there.

PhilPJL:
> I can't see this comment in the GitHub UI

Yes sorry I realized it was a patch and delete my comment.

SolastaCommunityExpansion/CustomFeatureDefinitions/CustomEffectForm.cs
ChrisPJohn: I've updated to use an abstract class. In practice it did make actual usage cleaner.

SolastaCommunityExpansion/Helpers/Extensions.cs
```diff
@@ -48,4 +48,25 @@ public static bool IsEnabled(this GameGadget gadget)
...
+        Motion = EffectForm.EffectFormType.Motion,
+        SpellSlots = EffectForm.EffectFormType.SpellSlots,
+        Divination = EffectForm.EffectFormType.Divination,
+        ItemProperty = EffectForm.EffectFormType.ItemProperty,
+        Alteration = EffectForm.EffectFormType.Alteration,
+        Topology = EffectForm.EffectFormType.Topology,
+        Revive = EffectForm.EffectFormType.Revive,
+        Kill = EffectForm.EffectFormType.Kill,
+        ShapeChange = EffectForm.EffectFormType.ShapeChange,
+        Custom = 9000,
```
PhilPJL, line 70: Is it better to go negative for additional enums?

PhilPJL (dismissed):
I would personally change AutoPreparedSpellsList to take IEnumerable<...> not an array

ChrisPJohn:
> I would personally change AutoPreparedSpellsList to take IEnumerable<...> not an array

If we want to change that it should be a separate change, that is not how this spell is going to be included, I was just adding it to that list for testing purposes.

SolastaCommunityExpansion/Helpers/Extensions.cs
ChrisPJohn: I used 9000 as the starting point to be consistent with what witch class was doing. I'm open to other suggestions.

ThyWoof (approved):

SolastaCommunityExpansion/Patches/CustomEffectForm/EffectDescriptionPatcher.cs
```diff
@@ -0,0 +1,20 @@
+﻿
+
+using System.Collections.Generic;
+using System.Linq;
+using HarmonyLib;
+
+namespace SolastaCommunityExpansion.Patches.CustomEffectForm
+{
+    [HarmonyPatch(typeof(EffectDescription), "FillTags")]
+    internal static class EffectDescriptionFillTags
```
ThyWoof, line 10: Not using our standard to name patches

SolastaCommunityExpansion/Patches/CustomEffectForm/RulesetImplementationManagerPatcher.cs
```diff
@@ -0,0 +1,24 @@
+﻿
+
+using System.Collections.Generic;
+using System.Linq;
+using HarmonyLib;
+
+namespace SolastaCommunityExpansion.Patches.CustomEffectForm
+{
+    [HarmonyPatch(typeof(RulesetImplementationManager), "ApplyEffectForms")]
+    internal static class RulesetImplementationManagerApplyCustomEffectForm
```
ThyWoof, line 10: Not using our standard to name patches

ThyWoof (commented):
All cool except for patch class names standardization

ThyWoof (approved):

ThyWoof:
merging this to start working on exposing all custom spells from a clear state

PR #294, opened 2021-12-24
Remove display of entrances - not useful

PhilPJL:


ThyWoof:
should we also block this code to only work in custom dungeons? @PhilPJL

PhilPJL:
Do you know how to do that?

ThyWoof:
Will approve and tweak. Easier

ThyWoof (approved):

PR #295, opened 2021-12-24
Late init

ThyWoof:


ThyWoof:
moving some initializations to happen after all mods are loaded. this should allow new blueprints from other mods to be affected by CE changes.

ThyWoof:
next step is to refactor Flex. Races to keep the mapping data as string instead of BPs references. This should also allow us to tweak unofficial races although not fully dynamic. Any new race / background out there would require this mapping to be updated. Considering moving over this mapping into settings but not firm on this one yet. If so will need a serializable dict as well.

PhilPJL (dismissed):

ThyWoof:
unfortunately we cannot move Feats / FightingStyles to late init as they don't get offered this way. Changed strategy to refresh initial feats on late init. this is enough to get them on unofficial races.

PhilPJL (approved):

PR #298, opened 2021-12-26
add support to unofficial races to date

ThyWoof:


SolastaCommunityExpansion/Models/FlexibleRacesContext.cs
```diff
@@ -59,40 +61,69 @@ private static void RemoveMatchingFeature(List<FeatureUnlockByLevel> unlocks, Fe
...
             foreach (var keyValuePair in removedFeatures)
             {
-                var exists = keyValuePair.Key.FeatureUnlocks.Exists(x => x.FeatureDefinition == keyValuePair.Value);
+                var characterClassDefinition = dbCharacterRaceDefinition.GetElement(keyValuePair.Key, true);
 
-                if (exists && enabled)
+                if (characterClassDefinition == null)
                 {
-                    RemoveMatchingFeature(keyValuePair.Key.FeatureUnlocks, keyValuePair.Value);
+                    Main.Log($"Race {keyValuePair.Key} not loaded.");
```
ChrisPJohn, line 99: Do we need to log if a race isn't available? This seems like it could be a common occurrence.

ChrisPJohn (approved):

PR #301, opened 2021-12-26
remove custom items before char. export

ThyWoof:
need to test


ThyWoof:
tested. seems to be working.

PhilPJL (approved):

PR #302, opened 2021-12-26
Classes UI support

ThyWoof:


ThyWoof:
added a hidden toggle to allow some initial support code to flow in and the initial block for classes in UI

SolastaCommunityExpansion/Viewers/Displays/SubClassesDisplay.cs
```diff
@@ -54,6 +54,41 @@ internal static void DisplaySubclasses()
                 MasterManipulator.UpdateSpellDCBoost();
             }
 
+            if (Main.Settings.EnableBetaFeaturesInMod)
+            {
+                //
+                // TODO: should we take same approach we do in subclasses? It isn't like we will have new classes every now and then...
```
ChrisPJohn, line 60: I prefer automating where possible. It makes keeping things like text descriptions up to date easier because they're only in 1 spot then as well as the nexus mod description. We don't need to automate it right now.

ThyWoof: I can do the automation but we would need to at least bring that abstract class code in this same request.

PR #304, opened 2021-12-27
UI to support new classes

ThyWoof:
this one is following the same pattern as subclasses, feats, etc

### DIFF
--- a/SolastaCommunityExpansion/Models/ClassesContext.cs
+++ b/SolastaCommunityExpansion/Models/ClassesContext.cs
@@ -1,0 +1,79 @@
+﻿using SolastaCommunityExpansion.Classes;
+using SolastaModApi.Extensions;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace SolastaCommunityExpansion.Models
+{
+    internal static class ClassesContext
+    {
+        public static Dictionary<string, AbstractClass> Classes { get; private set; } = new Dictionary<string, AbstractClass>();
+
+        internal static void Load()
+        {
+            //LoadClass(new Tinkerer());
+            //LoadClass(new Witch());
+        }
+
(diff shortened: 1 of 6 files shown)


ThyWoof:
had to closed previous one as messed up with a merge. followed your idea to make it dynamic @ChrisPJohn .

SolastaCommunityExpansion/Models/ClassesContext.cs
```diff
@@ -0,0 +1,79 @@
...
+using SolastaModApi.Extensions;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace SolastaCommunityExpansion.Models
+{
+    internal static class ClassesContext
+    {
+        public static Dictionary<string, AbstractClass> Classes { get; private set; } = new Dictionary<string, AbstractClass>();
```
ChrisPJohn, line 11: Why not make this <string, CharacterClassDefinition>, and then we can get rid of the AbstractClass and just need to have the class declaration provide the CharacterClassDefinition somehow.

ChrisPJohn (changes requested):
I think we can simplify things by just storing the class definition instead of an AbstractClass.

I did an abstract class for subclasses because it needed multiple pieces of information which the full classes don't.

SolastaCommunityExpansion/Models/ClassesContext.cs
ThyWoof: Because we're the kings of copy-n-paste ;-). Seriously I would rather keep as is so we can generalize the UI code at some point. Classes, Subclasses, Feats and Fighting Styles all use exactly the same logic. We should at some point a generic interface we can pass to an UI generic display method.

ChrisPJohn (approved):

PR #305, opened 2021-12-27
Refactor NegativeFeature so that it can be used outside of Thug

ChrisPJohn:
I'm not a fan of requiring the tag of the feature that's getting removed. That just feels like it's likely to break. The alternative of just searching every active feature though doesn't seem great either.


ThyWoof:
I need both otherwise this might break some MC combinations. What we can do instead is set the int level and feature we want removed and internally call the game method that creates the tag.

ThyWoof (approved):

ThyWoof:
Check the new request with a better interface.

PR #306, opened 2021-12-27
better interface for NegativeFeatureBuilder / support for subclasses

ThyWoof:


ThyWoof:
wrote these changes based on @ChrisPJohn's comments on last merge. It now has support for removal of features in sub classes and no more passing Tag strings. It's now encapsulated and using AttributeDefinitions static methods to derive.

ThyWoof:
I also changed the suffix on Thug from Remove to ThugRemove. This ensures others can remove same Rogue feature without name colide. Would it be better to pass the original class or subClass holding the removal feature instead of a name, guid tuple? Do the calculation inside by setting name = FeatureToRemove.Name + OwnerClassOrSubClass.Name + "Remove"?

SolastaCommunityExpansion/Subclasses/Rogue/Thug.cs
```diff
@@ -30,6 +30,7 @@ internal override CharacterSubclassDefinition GetSubclass()
 
         internal Thug()
         {
+            var featureName = AdditionalDamageRogueSneakAttack.Name + "ThugRemove";
```
ChrisPJohn, line 33: It's bad practice to rename features that are out in users' saved games. I understand that technically since this feature only does something at level up it's unlikely to have any negative effects on users (except error logs when they load), so we can rename if you think it's really important.

SolastaCommunityExpansion/CustomFeatureDefinitions/NegativeFeature.cs
```diff
@@ -3,23 +3,24 @@
...
         public FeatureDefinition FeatureToRemove;
+        public CharacterClassDefinition CharacterClass;
+        public CharacterSubclassDefinition CharacterSubclass;
+        public string Tag => CharacterSubclass == null ? AttributeDefinitions.GetClassTag(CharacterClass, ClassLevel) : AttributeDefinitions.GetSubclassTag(CharacterClass, ClassLevel, CharacterSubclass);
     }
 
     public sealed class NegativeFeatureBuilder : BaseDefinitionBuilder<NegativeFeatureDefinition>
     {
-        public NegativeFeatureBuilder(string name, string guid, string tag, FeatureDefinition featureToRemove)
+        public NegativeFeatureBuilder(string name, string guid, FeatureDefinition featureToRemove, int classLevel, CharacterClassDefinition characterClass, CharacterSubclassDefinition characterSubclass = null)
```
ChrisPJohn, line 17: So the benefit is you don't need to know how the tags are generated, but now you need to provide a lot more information.

I think we should add a comment explaining this is the level/class/subclass that the feature you're removing was granted.

ChrisPJohn (changes requested):

SolastaCommunityExpansion/CustomFeatureDefinitions/NegativeFeature.cs
ThyWoof: comment added and also renamed the Feature to a better naming convention.

SolastaCommunityExpansion/Subclasses/Rogue/Thug.cs
ThyWoof: good call and no need to rename. we should keep an eye though so in the future we adopt a better naming convention if we reuse this feature in other sub classes

ChrisPJohn (approved):

PR #309, opened 2021-12-28
move constants off settings into respective context namespace

ThyWoof:


SolastaCommunityExpansion/Models/EpicArrayContext.cs
```diff
@@ -2,6 +2,12 @@
 {
     internal static class EpicArrayContext
     {
+        internal const int GAME_MAX_ATTRIBUTE = 15;
+        internal const int GAME_BUY_POINTS = 27;
+
+        internal const int MOD_MAX_ATTRIBUTE = 17;
```
ChrisPJohn, line 8: By having this in settings we were allowing users to tweak the max attribute and points available if they wanted to. Is that functionality we want to have easily available to users?

(This question applies to most of these values),

ThyWoof: these values were never dumped into settings AFAIK. Only public fields.

ChrisPJohn (approved):

PR #310, opened 2021-12-28
add CTRL-SHIT-T hotkey to teleport party to cursor location

ThyWoof:
someone at Discord threw this idea. Code was ready from Encounters so why not... Will add UI / Settings later after the other requests get merged.


SolastaCommunityExpansion/Translations-en.txt
```diff
@@ -412,6 +412,8 @@ Message/&CharacterExportModalContentDescription	Enter the hero name to export:
 Message/&CharacterExportModalTitleDescription	Character Export
 Message/&SpawnCustomEncounterDescription	Spawn encounter at coordinates ({0}, {1})?
 Message/&SpawnCustomEncounterTitle	Encounter
+Message/&TeleportPartyDescription	Teleport party at coordinates ({0}, {1})?
```
ChrisPJohn, line 415: We should probably include a warning that this can cause quests/maps to break by bypassing triggers.

ThyWoof: yep. can do that. will wait for a fresh DEV branch first as also need to change some files that got changed on other requests. Not a big fan of resolving merging.

ThyWoof: done on both pop up and mod UI

SolastaCommunityExpansion/Viewers/Displays/ToolsDisplay.cs
```diff
@@ -30,6 +30,14 @@ internal static void DisplayTools()
                 Main.Settings.EnableTelemaCampaign = toggle;
             }
 
+            toggle = Main.Settings.EnableTeleportParty;
+            if (UI.Toggle("Enable the hotkey " + "ctrl-shift-(T)eleport".cyan() + " in game locations" + "\nYou might break quests or maps if you teleport to an undiscovered place".italic().yellow(), ref toggle))
```
ChrisPJohn, line 34: So it might be useful to ask a mod maker how best to word this. My understanding is that any teleporting (not just to somewhere undiscovered) could break quests/maps.

For example there could be a room with one door and an interactable item in that room. When the players interact with the item a trigger for going through the door could then be activated and the quest/map might only advance when characters go through the door again.

SolastaCommunityExpansion/Translations-en.txt
```diff
@@ -412,6 +412,8 @@ Message/&CharacterExportModalContentDescription	Enter the hero name to export:
 Message/&CharacterExportModalTitleDescription	Character Export
 Message/&SpawnCustomEncounterDescription	Spawn encounter at coordinates ({0}, {1})?
 Message/&SpawnCustomEncounterTitle	Encounter
+Message/&TeleportPartyDescription	Teleport party to coordinates ({0}, {1})?\nYou might break quests or maps if you teleport to an undiscovered place
```
ChrisPJohn, line 415: Same comment about how this wording should be run by a DM user.

ThyWoof: Got ok from Silverquick on warning message

vel8b8: LGTM

ChrisPJohn (approved):

PR #323, opened 2021-12-29
fix flexible races gnome

ThyWoof:


SolastaCommunityExpansion/Models/FlexibleRacesContext.cs
```diff
@@ -17,15 +17,15 @@ internal static class FlexibleRacesContext
             "FlexibleRaces/&PointPoolAbilityScore4Description",
             "FlexibleRaces/&PointPoolAbilityScore4Title");
 
-        private static readonly FeatureUnlockByLevel attributeChoiceFour = new FeatureUnlockByLevel(new FeatureDefinitionPointPoolBuilder("PointPoolAbilityScore4",
+        private static readonly FeatureUnlockByLevel _ = new FeatureUnlockByLevel(new FeatureDefinitionPointPoolBuilder("PointPoolAbilityScore4",
```
PhilPJL, line 20: Does this need to be kept for backward compatibility?  If so add comment.

SolastaCommunityExpansion/Models/FlexibleRacesContext.cs
```diff
@@ -47,7 +47,7 @@ internal static class FlexibleRacesContext
             { "HalfOrc", new List<string> { "FeatureSetHalfOrcAbilityScoreIncrease" } },
             // unofficial races
             { "FirbolgRace", new List<string> { "AttributeModifierFirbolgStrengthAbilityScoreIncrease", "AttributeModifierFirbolgWisdomAbilityScoreIncrease" } },
-            { "Gnome", new List<string> { "AttributeModifierGnomeAbilityScoreIncrease", "AttributeModifierForestGnomeAbilityScoreIncrease" } }
+            { "GnomeRace", new List<string> { "AttributeModifierGnomeAbilityScoreIncrease", "AttributeModifierForestGnomeAbilityScoreIncrease" } }
         };
 
         private static void RemoveMatchingFeature(List<FeatureUnlockByLevel> unlocks, FeatureDefinition toRemove)
```
PhilPJL, line 53: Although not part of this fix, this method would probably be better as:

```
        private static void RemoveMatchingFeature(List<FeatureUnlockByLevel> unlocks, FeatureDefinition toRemove)
        {
            unlocks.RemoveAll(f => f.FeatureDefinition.GUID == toRemove.GUID);
        }

```

SolastaCommunityExpansion/Models/FlexibleRacesContext.cs
ThyWoof: rolled back

ThyWoof:
half-elf is indeed 4...

PhilPJL (approved):

PR #326, opened 2021-12-29
fix teleporting party of 6

ThyWoof:


PhilPJL (approved):
Why didn't it work before?  Looks ok to me.

PR #327, opened 2021-12-30
adds support to only reveal gadgets on map after they are discovered

ThyWoof:
@PhilPJL , we cannot mess up too much with the ReferenceBox object in the gadget. Took this approach as this is the only place where we care some gadgets to respond with a valid ReferenceBox. You should now decide what to tweak in array `gadgetBlueprintsToRevealAfterDiscovery `.


ThyWoof:
@PhilPJL , added support for teleporter nodes (both individual and party although bot differentiating both now) and teleporter destinations. For now I reused the 2 icons we had and added TODO for them to be changed to something better.

ChrisPJohn (approved):
Looks good to me, but I think @PhilPJL is much more familiar with this code than I am, so if he's around he should take a look as well.

PR #332, opened 2021-12-30
fix on gathering user gadgets from user location object / add virtual…

ThyWoof:
… exit support


ThyWoof:
better approach to enumerate gadgets in user rooms

ThyWoof:
image
no teleports showing

PhilPJL (approved):
Only issue I can see is a teleporter with Invisible=true has an icon.  Can fix later.

PR #344, opened 2021-12-30
Gadget CheckIsEnabled bug fix

PhilPJL:
TA code appears buggy.  Maybe it's as designed but it doesn't work for us.

```
GameGadget
{ 
public bool CheckIsEnabled() => 
    this.CheckConditionName("Param_Enabled", true, false) 
    || this.CheckConditionName("Enabled", true, true);
}
```

which will always return true.

@ThyWoof please check your test dungeon still works.


ThyWoof (approved):

ThyWoof:
it worked as before. Was wondering if invisible teleporters or exits should be visible on the map after discovery.

PR #353, opened 2021-12-31
fixes after merging Witch

ThyWoof:


ThyWoof:
. Some code clean up after Witch merge
. Standardization of custom Feature Definition names
. Added toggle for Help Power

SolastaCommunityExpansion/CustomFeatureDefinitions/FeatureDefinitionAttackDisadvantageAgainstNonSource.cs
```diff
@@ -5,7 +5,7 @@ namespace SolastaCommunityExpansion.CustomFeatureDefinitions
     /// <summary>
     /// Imposes disadvantage when attacking anyone but the source of the specified condition.
     /// </summary>
-    public class AttackDisadvantageAgainstNonSource : FeatureDefinition, ICombatAffinityProvider
+    public class FeatureDefinitionAttackDisadvantageAgainstNonSource : FeatureDefinition, ICombatAffinityProvider
```
ChrisPJohn, line 8: So I think these renames are fine, since I don't think any new mods are using these outside of CE, but technically these are part of the public API of CE and so could be relied on by others.

SolastaCommunityExpansion/Models/PowersContext.cs
```diff
@@ -1,77 +1,84 @@
...
+            //SwitchHelpPower();
+        }
 
-            Powers[powerName].GuiPresentation.SetHidden(!active);
+        internal static void SwitchHelpPower()
+        {
+            var dbCharacterClassDefinition = DatabaseRepository.GetDatabase<CharacterClassDefinition>();
 
-            if (active)
+            if (Main.Settings.AddHelpActionToAllClasses)
```
ChrisPJohn, line 33: I know we only have help right now, but should we make this a list like the other groups, so it can scale?

I know Grapple is on our nice to have feature list. Any more would likely have to be home brew though.

ChrisPJohn (commented):
I'm not adding comments here about classes/subclasses/spell lists since we talked elsewhere about it.

SolastaCommunityExpansion/CustomFeatureDefinitions/FeatureDefinitionAttackDisadvantageAgainstNonSource.cs
ThyWoof: There is only one modder right now writing a new subclass for CE. Should be safe for now.

SolastaCommunityExpansion/Models/PowersContext.cs
ThyWoof: I'm still not sure about how big this one can grow. I can certainly revisit this after we get this initial fix in. Right now we have powers disabled anyways...

ThyWoof:
There is a small bug setting up the initial state of the expand all button; 
It still lacks the Switch method;
Will work on these on a 2nd request;

ThyWoof:
Initial select all state fixed, call for Switch ready from UI code
Switch missing last calls to add / remove spell to / from class / subclass

ThyWoof:
@ChrisPJohn , this should be good to go for spells selection.

ChrisPJohn:
It's not just about modders working in CE, but any mod could be relying on
this. For example Tinkerer uses the shared pool features. I don't think
anybody is currently using these outside of CE, but if they were it would
break their mod.

SolastaCommunityExpansion/Models/SpellsContext.cs
```diff
@@ -8,51 +7,367 @@ namespace SolastaCommunityExpansion.Models
 {
     internal static class SpellsContext
     {
-        public static Dictionary<string, SpellDefinition> Spells { get; private set; } = new Dictionary<string, SpellDefinition>();
+        internal class SpellRecord
+        {
+            internal SpellRecord(SpellDefinition spellDefinition, List<string> suggestedClasses, List<string> suggestedSubclasses)
```
ChrisPJohn, line 12: I still disagree with using class and subclass here and in the UI. It should just be about spell lists.

I think using spell lists directly will be clearer for users and make all of this code simpler.

ChrisPJohn (changes requested):
I think having the spells context and settings UI based on class/subclass makes everything more complicated and will confuse users.

It should just be about spell lists.

SolastaCommunityExpansion/Models/SpellsContext.cs
ThyWoof: Ok. Let me try that

SolastaCommunityExpansion/Models/GuiWrapperContext.cs
```diff
@@ -0,0 +1,46 @@
...
+        internal static void RecacheSpells() 
+        {
+            var guiWrapperService = ServiceRepository.GetService<IGuiWrapperService>();
+
+            if (guiWrapperService == null)
+            {
+                return;
+            }
+
+            var spellDefinitionsMap = AccessTools.Field(guiWrapperService.GetType(), "spellDefinitionsMap").GetValue(guiWrapperService) as Dictionary<string, GuiSpellDefinition>;
```
PhilPJL, line 17: Don't forget we have a helper in ModApi so you can write
`guiWrapperService.GetField<x,y>("spellDefinitionsMap")`

SolastaCommunityExpansion/Spells/BazouSpells.cs
```diff
@@ -1,27 +1,30 @@
...
         {
-
-            spells.Add(BuildEldritchOrb());
-            spells.Add(BuildFindFamiliar());
-            spells.Add(BuildFrenzy());
-            spells.Add(BuildMinorLifesteal());
-            spells.Add(BuildPetalStorm());
-            spells.Add(BuildProtectThreshold());
-
+            SpellsContext.RegisterSpell(BuildEldritchOrb(), new List<string> { "WarlockClass", "ClassWitch" });
```
PhilPJL, line 20: Would RegisterSpell benefit from a params parameter?
So you can write

`SpellsContext.RegisterSpell(BuildEldritchOrb(), "WarlockClass", "ClassWitch");`

etc

ThyWoof: Agreed and done.

ThyWoof:
@ChrisPJohn , indeed UI much cleaner now:

image

SolastaCommunityExpansion/Models/SpellsContext.cs
```diff
@@ -8,65 +7,179 @@ namespace SolastaCommunityExpansion.Models
...
-        public static Dictionary<string, SpellDefinition> Spells { get; private set; } = new Dictionary<string, SpellDefinition>();
-
-        internal static void Load()
+        internal class SpellRecord
         {
-            // Generate Spells here and fill the list
-            List<SpellDefinition> spells = new List<SpellDefinition>();
+            internal SpellRecord(SpellDefinition spellDefinition, List<string> suggestedClasses)
+            {
+                var dbCharacterClassDefinition = DatabaseRepository.GetDatabase<CharacterClassDefinition>();
```
ChrisPJohn, line 14: This should be using spell list defintions, not classes.

https://github.com/SolastaMods/SolastaBlueprints/tree/main/SpellListDefinition

SolastaCommunityExpansion/Models/SpellsContext.cs
```diff
@@ -8,65 +7,179 @@ namespace SolastaCommunityExpansion.Models
...
 
-            // Build Spells
-            BazouSpells.CreateSpells(spells);
+        internal static readonly Dictionary<SpellDefinition, List<string>> RegisteredSpells = new Dictionary<SpellDefinition, List<string>>();
 
-            // Use the list of Spells to get the settings and ui set up.
+        private static readonly List<CharacterClassDefinition> casterClasses = new List<CharacterClassDefinition>();
 
-            foreach (SpellDefinition spell in spells)
+        internal static List<CharacterClassDefinition> GetCasterClasses
```
ChrisPJohn, line 33: This should also be by spell list and should support if a subclass (ex Green Mage) defines their own spell list.

ThyWoof: ok

ChrisPJohn (dismissed):

ThyWoof:
some sub classes implement CastSpell instead of MagicAffinity... Added that case as well

ThyWoof:
image

ThyWoof:
image
3rd screenshot is a charm. now list looks correct.

ChrisPJohn (approved):

PR #357, opened 2022-01-01
Gadgets

PhilPJL:
[TestLocations.zip](https://github.com/ChrisPJohn/SolastaCommunityExpansion/files/7798232/TestLocations.zip)

Made a few changes:
1) improved camping icon
2) GameGadget.CheckIsEnabled bug fix reworked and is now a PostFix - not used by our code 
3) Changed GameLocationScreenMap_BindGadgets to only use 'IsEnabled() - our version of CheckIsEnabled()' for exits.  Other gadgets like teleporters are added to the map even if disabled (but not if invisible)
4) Implemented a fix for button activators.  Bug is that after first use button activators have to be clicked twice.  See GameGadget_SetCondition.  I need to check this for other activators.

There's another bug not sure how to fix.  Using the 'exits' test location press the 'enable' then 'disable' activators in sequence until all of the exits disappear.  The initially enabled exits don't disable until 2nd attempt.  Reason is two events get sent, one disable followed by an enable.  Eventually this extra event stops being sent.   The initially disabled  exits work ok.


ThyWoof (approved):

SolastaCommunityExpansion/Patches/GameUiScreenMap/GameGadgetPatcher.cs
```diff
@@ -78,57 +78,54 @@ internal static void Postfix(GameGadget __instance, ref bool ___revealed, ref bo
...
 
-            if(conditionIndex >= 0 && conditionIndex < ___conditionNames.Count)
+            if (conditionIndex >= 0 && conditionIndex < ___conditionNames.Count)
             {
                 var param = ___conditionNames[conditionIndex];
-                if (param == GameGadgetExtensions.Enabled || param == GameGadgetExtensions.ParamEnabled)
+
+                Main.Log($"GameGadget_SetCondition {__instance.UniqueNameId}: {param} state = {state}");
+
+                // BugFix? NOTE: not convinced this is correct thing to do.
```
ChrisPJohn, line 94: If this is a bug fix should it have it's own setting check? Should it be outside of the HideExitAndTeleporterGizmosIfNotDiscovered check?

ChrisPJohn (approved):
Approved with a question.

PR #359, opened 2022-01-02
hidden setting to display any unofficial feats

ThyWoof:


SolastaCommunityExpansion/Models/FeatsContext.cs
```diff
@@ -2,14 +2,28 @@
...
+
 namespace SolastaCommunityExpansion.Models
 {
     internal static class FeatsContext
     {
         public static Dictionary<string, FeatDefinition> Feats { get; private set; } = new Dictionary<string, FeatDefinition>();
 
+        internal static List<FeatDefinition> GetAllUnofficialFeats()
+        {
+            var officialFeatNames = typeof(SolastaModApi.DatabaseHelper.FeatDefinitions)
```
PhilPJL, line 18: Add a ToList to evaluate this query once, otherwise it will get run for each name in the .Where below

PhilPJL (approved):

PR #370, opened 2022-01-02
add a spell level filter to the spells UI

ThyWoof:


SolastaCommunityExpansion/Viewers/Displays/SpellsDisplay.cs
```diff
@@ -19,8 +23,18 @@ internal static class SpellsDisplay
 
         private static readonly Dictionary<string, bool> SpellNamesToggle = new Dictionary<string, bool>();
 
+        private static void RecacheSortedRegisteredSpells()
+        {
+            SortedRegisteredSpells.Clear();
```
PhilPJL, line 28: We've got a helper method SetRange which does Clear and then AddRange

PhilPJL (approved):

SolastaCommunityExpansion/Viewers/Displays/SpellsDisplay.cs
ThyWoof: ok. still working on this file. will change that on next.

PR #371, opened 2022-01-02
Circle of the Forest Guardian Druid Subclass

Dreadmaker:
Hey folks! This is my first mod/subclass, _and_ first project in C# in a _long_ time, so apologies in advance if it's on the rusty side.

This is my implementation of a homebrew subclass, the Circle of the Forest Guardian Druid. The goal is for it to be a melee gish druid, and be a frontline tank. Here's the `design document` : 

-------------------

**Circle of the Forest Guardian**

**Level 2**
**Expanded Spell List:**
- 2nd: Shield, Fog Cloud
- 3rd: Blur, Flame Blade
- 5th: Protection From Energy, Dispel Magic
- 7th: Flame Shield, Death Ward
- 9th: Hold Monster, Greater Restoration

**Sylvan Durability**
Forest Guardians naturally gain the resilience of a mighty oak. At 2nd level, your hit point maximum increases by 2 and increases by 1 whenever you gain a level in this class. 

**Sylvan War Magic**
You are used to wielding druidic magic in harmony with martial prowess. At 2nd level, you may cast spells that require a somatic component even if both of your hands are holding a weapon or shield.

**Bark Ward**
You gain the ability to summon a Bark Ward to protect you. As a bonus action, you can expend a use of your Wild Shape to create a Bark Ward, gaining 4 temporary hit points for each level you have in this class.

**Level 6**
**Extra Attack**
Starting at 6th level, you can attack twice, instead of once, whenever you take the Attack action on your turn.

**Level 10**
**Improved Bark Ward**
You've learned to channel the power of thorns into your Bark Ward. While your Bark Ward is active, when melee attackers within 5 feet of you hit you with an attack, they take 2d8 piercing damage.

**Level 14**
**Superior Bark Ward**
Your Bark Ward's damage increases to 3d8. Additionally, while it's active, you are immune to poison damage.

-------------------

All of this has been successfully implemented, and is _mostly_ bug free - the one thing that's annoying, and I think caused by the fact that Bark Ward is a Shared Pool power with no ability to define its own number of uses, when it hasn't been used, it actually shows that it only has one use, and if you mouse over it, it says that you've used it -1 times. It doesn't impact functionality, though, so It's just a visual annoyance. 

I'm sure there are ways this can be improved/made more efficient. Let me know, and I'm happy to change things around. Thanks so much to @ThyWoof and @DubhHerder for all the help over the past few days getting this up and running - it was a fun project!

Screenshots!

Level 2 description in-game:
CircleOfTheForestGuardianLevel2Features
 
Level 6+ description in-game:
CircleOfTheForestGuardianLevel6+Features

Example picture of Improved Bark Ward power in game: 
Improved Bark Ward

Finally, a shot of two different druids with normal Bark Ward and Superior Bark Ward, respectively. Note the temp HP levels correctly corresponding with their overall levels - level 2 in the first case and 14 in the second.
DifferentBarkWards


SolastaCommunityExpansion/Subclasses/Druid/CircleOfTheForestGuardian.cs
```diff
@@ -0,0 +1,463 @@
...
+        internal override CharacterSubclassDefinition GetSubclass()
+        {
+            if (Subclass == null)
+            {
+                Subclass = BuildAndAddSubclass();
+            }
+            return Subclass;
+        }
+
+        private const string DruidForestGuardianDruidSubclassName = "DruidForestGuardianDruidSubclass";
```
ThyWoof, line 28: might be a good call to prefix this name with DD. This way if by any remote chance TA releases the same content, yours won't break.

SolastaCommunityExpansion/Translations-en.txt
```diff
@@ -48,12 +48,18 @@ Condition/&ConditionDualFlurryApplyDescription	You are building momentum...
...
+Condition/&ConditionBarkWardTitle Bark Ward
+Condition/&ConditionBarkWardDescription Protected by the spirits of the forest.
 Condition/&ConditionMarkedByArcanistDescription	Afflicted by Arcanist's Mark.
 Condition/&ConditionMarkedByArcanistTitle	Arcanist's Mark
 Condition/&Frenzied	This creature is frenzied and will attack the closest creature.
 Condition/&FrenziedTitle	Frenzied
 Condition/&HelpActionDescription	This creature has been distracted.
 Condition/&HelpActionTitle	Distracted
+Condition/&ConditionImprovedBarkWardTitle Improved Bark Ward
+Condition/&ConditionImprovedBarkWardDescription the spirits of the forest cause 2d8 piercing damage to melee attackers.
```
ChrisPJohn, line 60: I'd expect the descriptions to be capitalized, but I wasn't sure if you had a reason to not do that.

SolastaCommunityExpansion/Subclasses/Druid/CircleOfTheForestGuardian.cs
```diff
@@ -0,0 +1,465 @@
...
+                Definition.SetRetaliateRangeCells(retaliationRange);
+                Definition.SetRetaliatePower(retaliationPower);
+                Definition.SetGuiPresentation(guiPresentation);
+                Definition.SetAncestryDefinesDamageType(false);
+            }
+        }
+
+        private sealed class FeatureDefinitionAutoPreparedSpellsBuilder : BaseDefinitionBuilder<FeatureDefinitionAutoPreparedSpells>
+        {
+            public FeatureDefinitionAutoPreparedSpellsBuilder(string name, string guid, List<FeatureDefinitionAutoPreparedSpells.AutoPreparedSpellsGroup> autospelllists,
```
ChrisPJohn, line 277: Maybe I missed it,. but I don't see what this is doing that: https://github.com/SolastaMods/SolastaModApi/blob/main/SolastaModApi/BuilderHelpers/FeatureDefinitionAutoPreparedSpellsBuilder.cs
doesn't support.

Since this works for you, no need to change, but I want to make sure I understand the shortcomings of the mod api builders.

ChrisPJohn (commented):
Possible change in translations and a question about auto prepared spells builder.

SolastaCommunityExpansion/Translations-en.txt
Dreadmaker: Great catch - solved it there and in a few other places.

SolastaCommunityExpansion/Subclasses/Druid/CircleOfTheForestGuardian.cs
Dreadmaker: This one was my bad - I was actually taking notes from the Arcanist subclass here and used what they did, but in fact you're right - you have what I need in the ModAPI method already. It required a little change (there's no constructor for it, like there was in my version), but not too bad, and I rather use what's already there. Thanks for the catch!

ThyWoof (approved):
Looks ok to me but let's wait on @ChrisPJohn final approval

PhilPJL (approved):

PR #374, opened 2022-01-03
UI colors

ThyWoof:


ChrisPJohn (approved):

SolastaCommunityExpansion/Models/FeatsContext.cs
```diff
@@ -86,7 +90,7 @@ public static string GenerateFeatsDescription()
 
             outString.Append("\n[list]");
 
-            foreach (var feat in Feats.Values)
+            foreach (var feat in Feats.Keys)
```
PhilPJL, line 93: Does this need to be ordered by FormatTitle()?

ThyWoof: Good call but as all others are like this won’t touch this one now

Issue #349, opened 2021-12-31
Use right click to cancel cursor


PhilPJL:
Allow cancellation of selected item with right mouse click.

![image](https://user-images.githubusercontent.com/271656/147818226-bc903617-81a7-4bab-b193-15b41e4bfa50.png)

PR #378, opened 2022-01-03
Cancel edit on right click

PhilPJL:
Allow cancellation of DM editor operation using mouse right click instead of pressing Esc - it's a pain on my keyboard :)


PhilPJL:
close #349

ThyWoof (approved):

PR #393, opened 2022-01-04
Upcast conjure elemental

PhilPJL:
Stage 1 done and seems to work ok.  Maybe don't merge unless Chris is happy with this approach.
Stage 2 is to add CR6 and 7 versions of the fire, air, earth elementals - need some stats


DubhHerder (approved):
Looks fine to me, especially if it's a quick implementation as noted. One thing that might cause confusion for some users is calling some of the elementals huge when they are still large size creature but that's really minor and doesn't effect implementation. I didn't look at the scaling in terms of balance etc

SolastaCommunityExpansion/Models/UpcastSummonsContext.cs
```diff
@@ -0,0 +1,305 @@
...
+            {
+                return new MonsterBuilder(name, CreateGuid(name), title, description, baseMonster);
+            }
+
+            string CreateGuid(string name) => GuidHelper.Create(Namespace, name).ToString("N");
+        }
+
+        internal static void ApplyUpcastSummon(EffectForm effectForm, int effectiveLevel)
+        {
+            if (!Main.Settings.EnableUpcastConjureElemental
```
ChrisPJohn, line 228: What's the plan for scaling this to support other summon spells (like fey as the comment states above)? I feel like either the setting should be more general (like EnableUpcastSummoningSpells) or there should be a check for for if the spell is conjure elemental.

SolastaCommunityExpansion/Models/UpcastSummonsContext.cs
```diff
@@ -0,0 +1,305 @@
...
+
+                if (effectiveLevel <= upcastSummonInfo.OriginalSpellLevel)
+                {
+                    Main.Log($"UpcastSummon-ApplySpellLevel: {effectiveLevel} <= {upcastSummonInfo.OriginalSpellLevel} - ignoring");
+                    return;
+                }
+
+                if (effectiveLevel - upcastSummonInfo.OriginalSpellLevel > upcastSummonInfo.UpcastMonsterDefinitionNames.Length)
+                {
+                    Main.Log($"UpcastSummon-ApplySpellLevel: {effectiveLevel}, no suitable upcast - ignoring");
```
ChrisPJohn, line 253: Should this use the highest available? So if the spell is 3rd level and there's upcasts for 4 and 5 but not 6, if I use a 6th level slot I should get the 5th level version.

ChrisPJohn (changes requested):

SolastaCommunityExpansion/Models/UpcastSummonsContext.cs
PhilPJL: I change this to just Elementals for now and then rethink when I see what TA finalize for Conjure Fey

SolastaCommunityExpansion/Models/UpcastSummonsContext.cs
PhilPJL: Good point.

PhilPJL: Done

PhilPJL:
Works as is, and I've bumped up the elemental to hit and damage bonus.

SolastaCommunityExpansion/Builders/MonsterAttackDefinitionBuilder.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        public MonsterAttackDefinitionBuilder SetToHitBonus(int value)
+        {
+            Definition.SetToHitBonus(value);
+            return this;
+        }
+
+        public MonsterAttackDefinitionBuilder SetDamageBonus(int value)
+        {
```
DubhHerder, line 20: This seems a bit targeted and specific for a builder, if it's included in the builders, maybe the name could mention that it only changes the first effect form damage bonus. It wouldn't work on monster attacks effect descriptions that use any other effect form  in the first entry or effect descriptions with multiple damage entries.

DubhHerder (commented):
Just a small note on the monster attack builders

SolastaCommunityExpansion/Builders/MonsterAttackDefinitionBuilder.cs
PhilPJL: I realised that :)  Need something like

```
        public MonsterAttackDefinitionBuilder SetDamageBonus(int value)
        {
            var effectForm = Definition.EffectDescription.GetFirstFormOfType(EffectForm.EffectFormType.Damage);
            effectForm?.DamageForm.SetBonusDamage(value);

            return this;
        }

```

PhilPJL: I meant to look at what you'd done but forgot.

DubhHerder: I did pretty much the same as yourself (copy an existing attack and alter it's parameters) just without the builder as i often wanted to change more than just the first damage bonus entry by adding new effect forms, changing the secondary damage bonuses, die types, damage types etc. 
If i was doing it again, i'd consider using Chris's approach of new effect descriptions etc via the general builders and effectdescription.copy(). 
For reference, here's where i defined monster attacks  but probably not much there you don't already know
https://github.com/ThyWoof/SolastaDungeonMakerPro/blob/master/SolastaDungeonMakerPro/Models/MonsterHelpers/NewMonsterAttacks.cs

PhilPJL: Ha, I don't know much, I've left all the new monsters, classes etc to Chris and you.

ChrisPJohn (approved):

PhilPJL:
Going for subspell filter.

PR #396, opened 2022-01-05
Update conjure elemental custom ef

PhilPJL:
First attempt at using custom effect form.  It works in that is summons a CR6 elemental, but for some reason it's hostile.  Same code on other branch works fine.


ChrisPJohn:
> First attempt at using custom effect form. It works in that is summons a CR6 elemental, but for some reason it's hostile. Same code on other branch works fine.

It sounds like the condition that makes the elemental mind controlled isn't getting applied:
https://github.com/SolastaMods/SolastaBlueprints/blob/main/SpellDefinition/ConjureElementalAir.3cd8a362519adfc4c87ddfa32298046d.json#L126

Looking through your code though I would expect it to be working.

PhilPJL:
It works now.  The SummonForm copy was incorrect.

ChrisPJohn (approved):

PhilPJL:
Going for subspell filter.

PR #403, opened 2022-01-06
Upcast conjure elemental using subspell filter

PhilPJL:
3rd version :)

Add additional sub-spells to ConjureElemental/Fey and filter the pop up based on slot level.

No up-cast
image

Up-cast
image

### DIFF
--- a/SolastaCommunityExpansion/Builders/MonsterAttackDefinitionBuilder.cs
+++ b/SolastaCommunityExpansion/Builders/MonsterAttackDefinitionBuilder.cs
@@ -1,0 +1,27 @@
+﻿using SolastaModApi;
+using SolastaModApi.Extensions;
+
+namespace SolastaCommunityExpansion.Builders
+{
+    public class MonsterAttackDefinitionBuilder : BaseDefinitionBuilder<MonsterAttackDefinition>
+    {
+        public MonsterAttackDefinitionBuilder(string name, string guid, MonsterAttackDefinition baseDefinition) :
+            base(baseDefinition, name, guid)
+        {
+        }
+
+        public MonsterAttackDefinitionBuilder SetToHitBonus(int value)
+        {
+            Definition.SetToHitBonus(value);
+            return this;
+        }
+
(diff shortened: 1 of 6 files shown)


ChrisPJohn:
I think this is my least favorite of the 3 in terms of user experience. I know we could improve that through additional patches, but I think the other two are better in terms of behaving more similar to how current spells upcast.

PhilPJL:
The reason I tried this way is because TA have implemented Conjure Fey as:
image
image

Conjure Elemental RAW works the same way so if we wanted them to match we would add lower level elementals to the sub-spell list, and higher level ones to be filtered by slot level.

ChrisPJohn:
> The reason I tried this way is because TA have implemented Conjure Fey as: image image
> 
> Conjure Elemental RAW works the same way so if we wanted them to match we would add lower level elementals to the sub-spell list, and higher level ones to be filtered by slot level.

I guess the part that's weird to me though is with your subspell implementation a user could accidentally summon the 5th level version when they're using a 6th level spell slot. What I meant by the other two behaving more similar to existing spells, is if you use a higher level slot to cast a spell you just get the benefit, you don't need to know to select the higher level slot and then the special higher level subspall.

PhilPJL:
The same thing with RAW and TAs implementation of conjure fey. You can use an L6 slot to conjure a dryad.
Doesn't make much sense in the game but maybe in tabletop. Does it, I'm not a tabletop player?

ChrisPJohn:
> The same thing with RAW and TAs implementation of conjure fey. You can use an L6 slot to conjure a dryad. Doesn't make much sense in the game but maybe in tabletop. Does it, I'm not a tabletop player?

That's a good point. I think I'm just used to new summon spells which are designed for scaling much better, I never played a character that used the conjuration spells so they aren't something I have any experience with. While clunky this may be the representation closest to the tabletop design, but it's a clunky design.

ChrisPJohn (approved):

PhilPJL:
I'll move this over to the beta branch and make some changes.

PhilPJL:
Closing for now - will move to beta branch

PR #407, opened 2022-01-07
Fix and tooltip

PhilPJL:
1) work around bug in GuiDropdown by using OptionDataAdvanced
2) add a tooltip
3) change to prefix for possible speed up

This PR could also go into Dev.


ThyWoof:
shouldn't we make this generic?

PhilPJL:
Finally got the drop down to resize to fit
image

ChrisPJohn (approved):

Issue #342, opened 2021-12-30
Add Higher level SRD spells


DubhHerder:
I didnt think it was worth opening individual issues for every single missing spell so here's a record of the higher level spells I've been working on, (will submit a PR when I've finished testing them) and the spells that will still need to be added. I've limited it to SRD spells as a core set.  

I've focused only on spells that are achievable without patching and maybe the reasons for not implementing certain spells with aid anyone attempting to add patches for the omitted spells. Of the implemented spells, the visuals might need reworking but the effects should be accurate. 

**7th Level** 
Primary reason for lack of 7th level spells is that the devs are more likely to do 7th level spells next so not a great use of time

|	Name		|	Status |	
|-----|-----|
|Arcane Sword          | no floating sword model available for the required effect proxy visual models (could reuse spiritual weapon)|
|Conjure Celestial     | no valid summons in game, would need to create a couatl monster but that would also require new at will powers to be accurate (could just give cleric spell casting as an approximation)                    |
|Delayed Blast Fireball| no ability to create trigger-able trap  via spells in base game and no timing feature to increase damage   |
|Divine Word           | don't think the effect description can handle multiple filtering requirements                                   |
|Etherealness          | haven't tried to implement has the base game code doesn't make use of existing enum movemode.Incorporeal |
|Finger of Death       | damage implemented , cant target both an individual to damage and a position to summon zombie in the same effect                                                                                           |
|Fire Storm            | effects cant target multiple AOE in a single spell , could implement as a stronger version of wall of fire with multiple shape options, rather than distinct effectproxy cubes                           |
|Forcecage             | wall of force not implemented so movement blocker possibly not working                                        |
|Magnificent Mansion   | N/A for combat game                                                              |
|Mirage Arcane         | N/A for combat game                                                                            |
|Plane Shift           | N/A for combat game                                                                               |
|Prismatic Spray       | game doesn't have capability variable effect via dice rolling                                                          |
|Project Image         | N/A for combat game                                                               |
|Regenerate            | avg healing of 40 points total makes the spell bad for 7th when you cannot lose body parts   |
|Resurrection          | already implemented by Devs                                                                      |
|Reverse Gravity       | Implemented, will submit PR after testing but fall damage for game seems to be limited to 1D6 no matter the height , same as when fly concentration is dropped                                                                            |
|Sequester             | N/A for combat game                                                                       |
|Simulacrum            | no mechanic to create double of unique player character/would need patch    |
[...]

PR #409, opened 2022-01-07
Adding SRD spells

DubhHerder:
For issue 342 (#342)

// 7th lvl
FingerOfDeath
ReverseGravity
ConjureCelestial

// 8th lvl
DominateMonster
Feeblemind
HolyAura
IncendiaryCloud
Maze
MindBlank
PowerWordStun
SunBurst

// 9th lvl
Foresight
MassHeal
MeteorSwarm_SingleTarget
PowerWordHeal
PowerWordKill
TimeStop
Shapechange
Weird

Note
PR does not have any toggle interaction with level 20 yet, will require another commit before merging for that

### DIFF
--- a/SolastaCommunityExpansion/Spells/SRDSpells.cs
+++ b/SolastaCommunityExpansion/Spells/SRDSpells.cs
@@ -3,6 +3,10 @@ using SolastaCommunityExpansion.CustomFeatureDefinitions;
 using SolastaCommunityExpansion.Models;
 using SolastaModApi;
 using SolastaModApi.Extensions;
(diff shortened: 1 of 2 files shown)


PhilPJL (commented):
My general comment is that the code doesn't conform to C# coding standards, like:
underscores shouldn't be used usually
local variables should be lower pascal case
etc.
But they can be fixed up later if necessary.

ThyWoof:
Agreed. I need a good sample to test so might end up merging this later today and start fixing names, etc

PR #413, opened 2022-01-08
only consider higher level spells if level 20 enabled

ThyWoof:


SolastaCommunityExpansion/Models/SpellsContext.cs
```diff
@@ -169,8 +170,19 @@ internal static void RegisterSpell(SpellDefinition spellDefinition, bool isFromO
 
             if (!RegisteredSpells.ContainsKey(spellDefinition))
             {
-                RegisteredSpells.Add(spellDefinition, new SpellRecord() { IsFromOtherMod = isFromOtherMod, SuggestedSpellLists = validateSpellLists });
-                RegisteredSpellsList.Add(spellDefinition);
+                if (spellDefinition.SpellLevel <= 6 || Main.Settings.EnableLevel20)
```
ChrisPJohn, line 173: I'm conflicted here. I know level 20 requires a restart to take effect, so there's no benefit to making enabling higher level spells possible without a restart. That said I do think some way to show what spells you'd have available at higher levels might be nice without having level 20 turned on.

Instead of this we could have a check where we add to spell lists and then have the UI do something like show a list of spell names that would be available with level 20 turned on. Thoughts?

PR #433, opened 2022-01-10
SRD Spells clean up / Add min. spell list set to UI

ThyWoof:


ThyWoof:
@ChrisPJohn : here are the changes I commit to beta. No need to focus on SRDSpells.cs for now. I had to do some clean up steps to pass over to @PhilPJL. Idea is he does the final clean up and then we review one last time for game correctness.

1) Now Spells UI always display higher level spells regardless of level 20. If level 20 not enable a warning gets displayed;
2) RegisterSpell interface takes a parameter with the default spell list I use now to create the minimum set;
3) Spells UI now has the min. set toggle
4) Changes on Translations were only to remove the _ to get back to standard

ChrisPJohn:
> @ChrisPJohn : here are the changes I commit to beta. No need to focus on SRDSpells.cs for now. I had to do some clean up steps to pass over to @PhilPJL. Idea is he does the final clean up and then we review one last time for game correctness.
> 
> 1. Now Spells UI always display higher level spells regardless of level 20. If level 20 not enable a warning gets displayed;
> 2. RegisterSpell interface takes a parameter with the default spell list I use now to create the minimum set;
> 3. Spells UI now has the min. set toggle
> 4. Changes on Translations were only to remove the _ to get back to standard

So making the change in two branches like this is going to cause merge issues. Generally the best practice is to make the change in once and then merge that in to the branches you want the change in. So while we have the beta branch around that would mean changes that are needed in both branches should get merged in to dev and then we merge dev into beta.

We can go ahead with this here now, but just something to be aware of to keep merging simpler in the future.

ThyWoof:
Good to know. I need to setup 2 game instances to properly do that

ChrisPJohn (approved):
I didn't review the new spells for correctness since I believe @DubhHerder tested them pretty thoroughly.

DubhHerder:
I can't really say whether the spells are correct or not anymore as there have been changes to the code beyond capitalization, removing underscores etc and I don't know if Zappa has tested those changes or not

PR #448, opened 2022-01-11
Beta fix save by location

PhilPJL:
Small tidy in `SubspellSelectionModal_OnActivate`

Fix issue in Save By Location.  Calling unnecessary `ValueChanged(guiDropdown);` causes items to be added twice.

This issue also exists in Dev where items are added 3 times because we're using a postfix.  Suggest not fixing dev.


ThyWoof:
What are the chances we loose these next time we merge dev here?

ThyWoof (approved):

PR #452, opened 2022-01-12
reorg sorting

ThyWoof:
leverages Classes and Subclasses contexts to reorder Features in respective databases;
added correct hidden setting to control sub classes sorting on UI;
removed 4 harmony patches as consequence (improved performance too as no more re-sorting these collections every time);


ThyWoof:
It also fix the features ordering on class selection (for example, witch maledictions

ChrisPJohn (approved):

PR #461, opened 2022-01-12
fix drag to ground when in char pool hero inspection

ThyWoof:
official and beta still have a null exception bug if you try to drag an item on ground when in char. pool inspection screen.


ThyWoof:
this is game code with the additional null check in place

PR #462, opened 2022-01-12
Fix battle timescale

ThyWoof:
using a Postfix instead. Still taking into account battles only as well as multiplayer scenarios.

### DIFF
--- a/SolastaCommunityExpansion/Patches/GameUiBattle/GameTimeSetTimeScalePatcher.cs
+++ b/SolastaCommunityExpansion/Patches/GameUiBattle/GameTimeSetTimeScalePatcher.cs
@@ -1,6 +1,5 @@
 ﻿using System.Diagnostics.CodeAnalysis;
 using HarmonyLib;
-using UnityEngine;
 
 namespace SolastaCommunityExpansion.Patches.GameUiBattle
 {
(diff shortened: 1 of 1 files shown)


ThyWoof:
need a campaign save file to fully test...

PhilPJL (commented):
I assume this is ok, but not really sure how to test.

SolastaCommunityExpansion/Patches/GameUiBattle/GameTimeSetTimeScalePatcher.cs
```diff
@@ -1,32 +1,24 @@
...
-            }
-            else
-            {
-                Time.timeScale = ___timeScale * (___fasterTimeMode ? Main.Settings.BattleCustomTimeScale : 1f);
-            }
-
-            return false;
+            ___timeScale = ___networkTimeScale != 1.0 
+                ? ___networkTimeScale 
+                : ___timeScale * (___fasterTimeMode || Main.Settings.PermanentlySpeedBattleUp ? Main.Settings.BattleCustomTimeScale : 1f);
```
PhilPJL, line 21: If fasterTimeMode == true (where does this get set?) the timeScale is already x 1.5 and now we're speeding up more, is that correct?

ThyWoof: Good catch. Should only consider the slider value

ThyWoof:
> I assume this is ok, but not really sure how to test.
> If it didn't work previously how does this change things?

This is code that was written way before Sorcerer DLC. I took the latest implementation that now also considers MP. Making a postfix also simplifies the logic to only trigger in battles.

PhilPJL (approved):

PR #495, opened 2022-01-16
remove virtual exits from logic

ThyWoof:
virtual exits only have standard highlight but no gizmos. removed from collection.


ThyWoof:
```        internal static readonly GadgetBlueprint[] GadgetBlueprintsWithGizmos = new GadgetBlueprint[]
        {
            Exit,
            ExitMultiple,
            TeleporterIndividual,
            TeleporterParty,
        };```controls which gadgets should be considered in a special way because of the gizmos

PhilPJL:
I'm going to test this in DM later

PhilPJL (approved):

PR #504, opened 2022-01-17
Add ctrl-h to show/hide all panels

PhilPJL:


ThyWoof:
@PhilPJL : shouldn't we add the same patch to GameLocationScreenBattle? Copy and paste from GameLocationScreenExploration renaming the class.

SolastaCommunityExpansion/Patches/GameUiLocation/GameLocationScreenExplorationPatcher.cs
```diff
@@ -23,56 +23,88 @@ internal static class GameLocationScreenExploration_HandleInput
...
-                    else
+                    ShowPartyControlPanel();
+                    break;
+
+                case GameUiContext.CTRL_H:
+                    ShowAll();
+                    break;
+            }
+
+            void ShowAll()
```
ChrisPJohn, line 46: The way this is currently implemented it looks like you could get into a weird state of showing half and hiding half of the panels by hitting ctrl+h. Should it be able to specify all hide/all show?

PhilPJL: Good catch.

PhilPJL:
> @PhilPJL : shouldn't we add the same patch to GameLocationScreenBattle? Copy and paste from GameLocationScreenExploration renaming the class.

Done.

> In that case I would move the entire logic to the Context and simply call the same method from both.

Not done yet.

ThyWoof (approved):

PR #507, opened 2022-01-17
refactor hotkeys in it's own file

ThyWoof:


ThyWoof:
refactored all hotkeys in it's own file after speaking with @PhilPJL to avoid future issues on assigning new ones.

PhilPJL (approved):

PR #509, opened 2022-01-18
fix twinned spells offering to prevent certain spells from being used…

ThyWoof:
… as official SRD rules


ThyWoof:
This patch has been in use for months in MC but I believe CE is a best place for it now that we're getting new spells that might need to be added to these lists.

ChrisPJohn (approved):

PR #531, opened 2022-01-20
Refactor follow chars

ThyWoof:
Patched FunctorTeleport directly using a transpiler. 
This is the functor that gets called by Gadget.Update() so it follows same idea as original.
Added same Postfix code to be called after TeleportCharacter call in FunctorTeleport.
1 less Prefix and Postfix in CE mod.


ChrisPJohn (approved):
This looks fine to me, but I also haven't been keeping up to date with the other teleport/follow work (e.g. don't follow character in battle)

PR #554, opened 2022-01-22
General cleanup using Roslynator.

PhilPJL:
And manually move some Tinkerer code around to allow use of static readonly.

Please note there are several of these 
#pragma warning disable S1481 // Unused local variables should be removed
that need sorting

Also there are few files I'm confused about.  For example GameUiContext is showing the addition of 
```
                   case Hotkeys.CTRL_SHIFT_H:
                        GameHud.ShowAll(gameLocationBaseScreen, GetInitiativeOrPartyPanel(), GetTimeAndNavigationPanel());
                        return;
```

which appears to have come from dev.  

The file ArtilleryConstruct_Force has changes I don't remember making.

### DIFF
--- a/SolastaCommunityExpansion/Builders/CastSpellBuilder.cs
+++ b/SolastaCommunityExpansion/Builders/CastSpellBuilder.cs
@@ -269,7 +269,6 @@ namespace SolastaCommunityExpansion.Builders
                 case CasterProgression.THIRD_CASTER:
                     for (; level < 21; level++)
                     {
-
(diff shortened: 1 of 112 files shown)


DubhHerder:
> The file ArtilleryConstruct_Force has changes I don't remember making.

I think you were adding back in code that was taken back out in https://github.com/ChrisPJohn/SolastaCommunityExpansion/pull/550/files

PhilPJL:
I've closed this. Will have to start again but don't have the enthusiasm right now.

PR #559, opened 2022-01-23
Beta cleanup

PhilPJL:
Pass 1:
Turn off warnings as errors for now until cleanup complete
Remove blank lines
Remove empty doc comments
Add static to class where necessary
Add const where necessary

Pass 2:
Add parentheses for clarity
Call method as extension method
Make builder instance properties readonly (easy ones)
Remove empty initializer (change {} to ())
Remove redundant .ToString()
Remove redundant parentheses
Rename local variables which hide fields with same name
Remove unnecessary interpolated string

Pass 3:
Optimize some LINQ calls
Simplify lazy initialization
Simplify logical negation
Use expression bodied lambda
Use readonly auto implemented property

Pass 4:
Use auto-implemented property
Use Count==0 instead of .Any()
User array index instead of ElementAt
Remove unnecessary null check
Avoid null refs

Pass 5:
Run VS cleanup - reorders usings, applies editor config preferences

Pass 6:
Simplify member access

### DIFF
--- a/SolastaCommunityExpansion/.editorconfig
+++ b/SolastaCommunityExpansion/.editorconfig
@@ -143,6 +143,12 @@ dotnet_diagnostic.S125.severity = silent
 # S3267: Loops should be simplified with "LINQ" expressions
 dotnet_diagnostic.S3267.severity = silent
 
+# RCS1146: Use conditional access.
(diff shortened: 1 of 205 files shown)


ThyWoof (approved):

ThyWoof:
created a betaBackup branch before this merge

PR #562, opened 2022-01-23
Final cleanup.

PhilPJL:
Apply previous cleanups to DM Pro code.
Change Tinkerer identifiers to pascal case.
Turn warnings as errors back on.

Problem with Roslynator warning ROS0003 - won't go away and can't disable it.  My problem only :)


SolastaCommunityExpansion/Classes/Tinkerer/ArtilleristBuilder.cs
```diff
@@ -237,10 +237,13 @@ public static CharacterSubclassDefinition Build(CharacterClassDefinition artific
             fire15Effect.SetParticleEffectParameters(DatabaseHelper.SpellDefinitions.BurningHands.EffectDescription.EffectParticleParameters);
 
             // TODO- add an option to enable the power version of the Blaster (there have been some requests for this) instead of the summons
+            // TODO: @Chris - review
```
ChrisPJohn, line 240: There's a todo literally the line above here about why this is here and what to do about it. This applies to all of these new todos in this file.

PhilPJL: ok, just damping down the warning

ThyWoof (approved):

PR #566, opened 2022-01-24
tweak adventure log feature on 1.3.10

ThyWoof:
no need to track Lore anymore. others still relevant. did some clean up on related files.


ThyWoof:
Also added another small reorg on 2 patches from level20 to DM

PhilPJL (approved):

PR #572, opened 2022-01-27
Patches refactor and bug fixes

ThyWoof:
. refactored 2 patches using a prefix / postfix pattern into one transpiler patch trapping the specific call
. added a bug fix when for some interactions on inventory panel @ main screen as there is no gameLocationCharacter
. split a patch file in 2


SolastaCommunityExpansion/Patches/SrdAndHouseRules/InventoryShortcutsPanelPatcher.cs
```diff
@@ -9,64 +11,60 @@ namespace SolastaCommunityExpansion.Patches.SrdAndHouseRules
...
     {
-        public static bool InOnCastLightCb { get; private set; }
-
-        public static void Prefix()
-        {
-            InOnCastLightCb = true;
-        }
-
-        public static void Postfix()
+        public static bool MyTryFindTargetWieldedItem(RulesetCharacterHero rulesetCharacterHero, out RulesetItem targetItem)
```
ThyWoof, line 14: need to be public or change line 55, add non public binding flags and add #pragma directives around that. simpler this way...

ThyWoof:
Also refactored patch from #573

PhilPJL (changes requested):
In testing quick cast light cantrip has reverted to using the wielded items by default.

PR #576, opened 2022-01-27
Check there's a translation entry for titles and descriptions

PhilPJL:


SolastaCommunityExpansion/Models/GuiPresentationCheck.cs
```diff
@@ -68,13 +69,36 @@ public static void PostCELoadCheck()
 
             var allDefinitions = GetAllDefinitions();
 
+            var languageSourceData = LocalizationManager.Sources[0];
+            var languageIndex = languageSourceData.GetLanguageIndexFromCode("en");
```
ThyWoof, line 73: can we make this generic? `LocalizationManager.CurrentLanguage`...

PR #592, opened 2022-01-31
UI from beta

ThyWoof:
. bring latest UI enhancements from Beta
. bring level map fix bug from Beta
. update to support C# 8 lang.
. migrate inventory panel bug fix from Beta [this was previously on MC which got removed on 1.2.15.18 released this morning]


ThyWoof:
Except for a toggle under SRD _ didn’t migrate conjure fey patch _ and the DMPRO merge, code under Viewers is almost identical to beta.

ChrisPJohn (approved):

PR #594, opened 2022-01-31
fix toggle camera

ThyWoof:


SolastaCommunityExpansion/Models/GameUiContext.cs
```diff
@@ -110,8 +110,10 @@ void ToggleZoomCamera()
                 }
                 else
                 {
-                    cameraService.DebugCameraEnabled = EnableDebugCamera;
+                    EnableDebugCamera = !EnableDebugCamera;
                 }
+
+                cameraService.DebugCameraEnabled = EnableDebugCamera;
```
ChrisPJohn, line 116: cameraService might be null

PhilPJL (approved):

PhilPJL (changes requested):
Oops, should have read Chris's comment

SolastaCommunityExpansion/Models/GameUiContext.cs
ThyWoof: fixed

SolastaCommunityExpansion/Models/GameUiContext.cs
```diff
@@ -110,8 +110,9 @@ void ToggleZoomCamera()
                 }
                 else
                 {
+                    EnableDebugCamera = !EnableDebugCamera;
                     cameraService.DebugCameraEnabled = EnableDebugCamera;
-                }
+                }
```
PhilPJL, line 115: I think this should be 

```
...
else { EnableDebugCamera = !EnableDebugCamera; }
if(cameraService != null) { camera ... }
```

PhilPJL (changes requested):
see comment

SolastaCommunityExpansion/Models/GameUiContext.cs
PhilPJL: Although what function does IViewService have here?

ThyWoof: I copied this code from game and added the flag logic. Really don't know what role IViewService has in here.

PhilPJL: Don't trust the game code :)

ThyWoof: done

PhilPJL (approved):

PR #604, opened 2022-01-31
fix db helpers generator

ThyWoof:


ChrisPJohn (approved):
I think this is fine. @PhilPJL should definitely take a look at this even if it has already been merged.

PhilPJL (approved):

PR #606, opened 2022-02-01
Swapped Title and Description

RealBazou:


ChrisPJohn (approved):
I wish I remembered why I put Description first when I originally made that builder. In hindsight it was probably the wrong decision, but at this point I don't think we can switch it easily.

PR #612, opened 2022-02-02
reorg builders and custom features

ThyWoof:
. initial reorg of Builders, Features and CustomFeatures;
. no code change except for new namespace names and corresponding using statements;

planned on step 2:

. split files under CustomFeatureDefinitions and get class that is a builder as a new file under Builders

### DIFF
--- a/SolastaCommunityExpansion/Api/AdditionalExtensions/GameGadgetExtensions.cs
+++ b/SolastaCommunityExpansion/Api/AdditionalExtensions/GameGadgetExtensions.cs
@@ -3,26 +3,8 @@ using System.Linq;
 using System.Reflection;
 using SolastaModApi.Infrastructure;
 
-namespace SolastaCommunityExpansion.Helpers
+namespace SolastaModApi.Extensions
 {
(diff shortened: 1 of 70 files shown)


ChrisPJohn (commented):
What's the intended distinction between files in Builders and Builders/Features? Is it just if the entity that is being built a Feature?

There used to be a distinction in completeness of a builder where the ones included in the ModAPI were more or less complete and generic (at least at the time they were added to ModAPI), while those in CE were often just done enough to support the feature that was being added. Do we want to try to keep track of that kind of code state/quality somewhere?

PhilPJL (approved):
I expect given Chris's comment we'll think about how builders are organized re
FeatureDefinition
OtherDefinition
Not a definition
Complete vs not complete?  I don't think builders should be organized like this.  Maybe a comment would be sufficient?

For example CastSpellBuilder is a FeatureDefinitionCastSpell and should be in Features at the moment.

Builders should all be named <class>Builder which affects at least CastSpellBuilder, SpellBuilder, SpellListBuilder.

Approving since this is Step1 of 'n'.

PR #623, opened 2022-02-03
reorg witch subclasses to be under Witch

ThyWoof:


ThyWoof:
Will apply same design later under Tinkerer

SolastaCommunityExpansion/SolastaCommunityExpansion.csproj
```diff
@@ -202,4 +202,8 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
+
+  <ItemGroup>
```
PhilPJL, line 206: I don't think this is necessary. There are no includes for other folders.

PhilPJL (changes requested):
Folder in proj file?

PhilPJL (approved):

PR #637, opened 2022-02-06
Merge Multiclass into CE

ThyWoof:
this code has been quite stable since Primal DLC. I also took the precaution of protecting every single patch with a global setting in case we need to fully disable it.

from here @PhilPJL can start doing some code clean up.

I also kept the global settings if checks separate on purpose. Please don’t consolidate them on clean ups.

### DIFF
--- a/SolastaCommunityExpansion/Multiclass/CustomDefinitions/ArmorProficiencies.cs
+++ b/SolastaCommunityExpansion/Multiclass/CustomDefinitions/ArmorProficiencies.cs
@@ -1,0 +1,94 @@
+﻿using System.Collections.Generic;
+using SolastaModApi;
+using SolastaModApi.Infrastructure;
+using static SolastaModApi.DatabaseHelper.FeatureDefinitionProficiencys;
+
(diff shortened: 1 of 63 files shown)


SolastaCommunityExpansion/Multiclass/CustomDefinitions/ArmorProficiencies.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System.Collections.Generic;
+using SolastaModApi;
+using SolastaModApi.Infrastructure;
+using static SolastaModApi.DatabaseHelper.FeatureDefinitionProficiencys;
+
+namespace SolastaCommunityExpansion.Multiclass.CustomDefinitions
+{
+    //internal static class ArmorProficiencies
```
ThyWoof, line 8: @PhilPJL : can you please check this code later? I can't see the added proficiencies if I take this approach so rolled back to what I had before.

PhilPJL (approved):

PR #669, opened 2022-02-07
better check on dmpro blueprints creation

ThyWoof:


SolastaCommunityExpansion/Models/DmProEditorContext.cs
```diff
@@ -193,7 +193,10 @@ private static void UnleashGadgetsOnAllEnvironments()
                         newGadgetBlueprint.PrefabsByEnvironment.Add(myPrefabByEnvironment);
                     }
 
-                    newGadgets.Add(newGadgetBlueprint);
+                    if (!newGadgets.Exists(x => x.Name == newGadgetBlueprint.Name))
```
PhilPJL, line 196: You can use TryAdd()

ThyWoof:
added 2 tweaks on MC as well. Reason why GUI patches cannot be protected is because we change the terms. If we protect them and MC disabled the tooltip on hero will produce a NULL as it expects 3 parameters with this mod.

PhilPJL (approved):

PR #673, opened 2022-02-08
refactor upcast conjured elemental or fey

ThyWoof:
. Feature now patches 2 methods instead of 3;
. Improved performance as `SpellDefinition.SubspellsList` patch was getting called 3 times from `SubSpellSelectionModal.Bind`. Now it gets cached once and reused;


PhilPJL (changes requested):
Doesn't work :)

ThyWoof:
@PhilPJL : couldn't open the shared save file but revisited the logic and indeed it was incorrectly filtering spells so I hope this change fixes it. please review with your save.

ChrisPJohn (approved):
The changes look fine. I assume they were verified in game.

ThyWoof:
It’s now working after I got the save file from @PhilPJL. Shared some screenshots as well but prefer him to do the final merge

PhilPJL (approved):
Works correctly now

PR #675, opened 2022-02-08
CTRL-click only swaps main hand

ThyWoof:
CTRL-click when swapping weapon configurations will only swap main hand. This is in case you have a good shield and wanna only swap weapons keeping the same shield in hands. What really ends up happening is the off hand item moves from one wielded config to another...


ThyWoof:
also add a setting to unleash 3rd row to be used with any items. There is still a refresh issue when the inventory first displays though...

PhilPJL (approved):
There is an issue with enabling the tertiary slot when both hands in use on selected slot.
Will create repro.

PR #692, opened 2022-02-09
refactor ApplySummonForm patch

ThyWoof:
remove patches from InflictCondition and RemoveCondition;
embed logic within ApplySummonForm patch using a transpiler;

### DIFF
--- a/SolastaCommunityExpansion/Patches/CustomFeatures/RulesetActorPatcher.cs
+++ b/SolastaCommunityExpansion/Patches/CustomFeatures/RulesetActorPatcher.cs
@@ -50,34 +50,6 @@ namespace SolastaCommunityExpansion.Patches.CustomFeatures
         }
     }
 
-    [HarmonyPatch(typeof(RulesetActor), "InflictCondition")]
-    [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
-
-    internal static class RulesetActor_InflictCondition
-    {
-        internal static void Prefix(string conditionDefinitionName,
-            ref int sourceAmount)
-        {
-            if (RulesetImplementationManagerLocation_ApplySummonForm.ConditionToAmount.ContainsKey(conditionDefinitionName))
-            {
-                sourceAmount = RulesetImplementationManagerLocation_ApplySummonForm.ConditionToAmount[conditionDefinitionName];
-            }
-        }
-    }
-
-    [HarmonyPatch(typeof(RulesetActor), "RemoveCondition")]
-    [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
-    internal static class RulesetActor_RemoveCondition
-    {
-        internal static void Postfix(RulesetActor __instance, RulesetCondition rulesetCondition)
-        {
-            if (rulesetCondition?.ConditionDefinition is INotifyConditionRemoval notifiedDefinition)
-            {
-                notifiedDefinition.AfterConditionRemoved(__instance, rulesetCondition);
-            }
-        }
-    }
-
     [HarmonyPatch(typeof(RulesetActor), "RollDie")]
     [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
     internal static class RulesetActor_RollDie
--- a/SolastaCommunityExpansion/Patches/CustomFeatures/RulesetImplementationManagerLocationPatcher.cs
+++ b/SolastaCommunityExpansion/Patches/CustomFeatures/RulesetImplementationManagerLocationPatcher.cs
@@ -1,7 +1,9 @@
 ﻿using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using System.Reflection.Emit;
 using HarmonyLib;
+using SolastaCommunityExpansion.CustomFeatureDefinitions;
 using SolastaModApi.Extensions;
 
 namespace SolastaCommunityExpansion.Patches.CustomFeatures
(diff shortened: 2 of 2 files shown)


SolastaCommunityExpansion/Patches/CustomFeatures/RulesetActorPatcher.cs
```diff
@@ -50,34 +50,6 @@ internal static void Postfix(RulesetActor __instance, RuleDefinitions.TurnOccure
...
-        }
-    }
-
-    [HarmonyPatch(typeof(RulesetActor), "RemoveCondition")]
-    [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
-    internal static class RulesetActor_RemoveCondition
-    {
-        internal static void Postfix(RulesetActor __instance, RulesetCondition rulesetCondition)
-        {
-            if (rulesetCondition?.ConditionDefinition is INotifyConditionRemoval notifiedDefinition)
```
ChrisPJohn, line 74: This is unrelated to the applySummon. I'm pretty sure your changes don't keep the feature working. This is used by the Path of Light Barbarian subclass.

ChrisPJohn (changes requested):

SolastaCommunityExpansion/Patches/CustomFeatures/RulesetActorPatcher.cs
ChrisPJohn: Would it be better if patches were grouped by feature they're enabling instead of throwing a whole bunch of patches in a file because the patch the same class?

ThyWoof: Thanks for reviewing. Will fix that.

This is exactly the idea on the updates we've been pushing over the last weeks. CustomFeatures is the last folder under patches that need a better organization.

Check, for example, what is under SrdRules. This is the goal.

SolastaCommunityExpansion/Patches/CustomFeatures/RulesetImplementationManagerLocationPatcher.cs
```diff
@@ -11,44 +12,69 @@ namespace SolastaCommunityExpansion.Patches.CustomFeatures
...
                     foreach (FeatureDefinitionSummoningAffinity summoningAffinity in formsParams.sourceCharacter.FeaturesToBrowse.OfType<FeatureDefinitionSummoningAffinity>())
                     {
-                        if (string.IsNullOrEmpty(summoningAffinity.RequiredMonsterTag) || element.CreatureTags.Contains(summoningAffinity.RequiredMonsterTag))
+                        if (string.IsNullOrEmpty(summoningAffinity.RequiredMonsterTag) || monsterDefinition.CreatureTags.Contains(summoningAffinity.RequiredMonsterTag))
                         {
                             foreach (ConditionDefinition addedCondition in summoningAffinity.AddedConditions)
                             {
                                 int sourceAmount = 0;
+
                                 switch (addedCondition.AmountOrigin)
```
ChrisPJohn, line 58: If we're going to use a transpiler I wonder if there's a better way to insert this switch into the original function so that we don't need to do all of the loops/checks above here twice. That would also allow us to get rid of the extendInflictCondition and the ConditionToAmount dictionary.

ChrisPJohn (changes requested):

SolastaCommunityExpansion/Patches/CustomFeatures/RulesetImplementationManagerLocationPatcher.cs
ThyWoof: moved switch to trigger inside ExtendedInflictCondition. This was the best place to patch as it's usually easier to path method calls and this happens right after the switch.

Please test @ChrisPJohn

ThyWoof:
Keep in mind we need to keep this under our radar before PRD release. Reason is dev build doesn’t have any optimization and big changes that local won’t be 35 anymore…

SolastaCommunityExpansion/Patches/CustomFeatures/RulesetImplementationManagerLocationPatcher.cs
```diff
@@ -11,62 +12,71 @@ namespace SolastaCommunityExpansion.Patches.CustomFeatures
...
+            int effectLevel,
+            string effectDefinitionName,
+            int sourceAmount,
+            int sourceAbilityBonus,
+            RulesetImplementationDefinitions.ApplyFormsParams formsParams,
+            ConditionDefinition addedCondition)
         {
-            SummonForm summonForm = effectForm.SummonForm;
-            if (summonForm.SummonType == SummonForm.Type.Creature && !string.IsNullOrEmpty(summonForm.MonsterDefinitionName) && DatabaseRepository.GetDatabase<MonsterDefinition>().HasElement(summonForm.MonsterDefinitionName))
+            switch (addedCondition.AmountOrigin)
```
ChrisPJohn, line 31: When I summon a witch's familiar (granted at level 2) a null pointer gets thrown here. I assume addedCondition is null.

ChrisPJohn (changes requested):

SolastaCommunityExpansion/Patches/CustomFeatures/RulesetImplementationManagerLocationPatcher.cs
ThyWoof: null pointer fixed but there is another issue on summoning a witch's familiar not related to this mod. I get this error on beta as well:

```
Invalid Sprite Reference in monster WitchOwl (MonsterDefinition)
UnityEngine.StackTraceUtility:ExtractStackTrace ()
UnityEngine.DebugLogHandler:LogFormat (UnityEngine.LogType,UnityEngine.Object,string,object[])
UnityEngine.Logger:LogFormat (UnityEngine.LogType,string,object[])
UnityEngine.Debug:AssertFormat (bool,string,object[])
Trace:Assert (bool,string,object[])
GuiCharacter:AssignPortraitImage (UnityEngine.UI.RawImage,UnityEngine.RectTransform,UnityEngine.RectTransform,System.Action)
CharacterPlateBase:CommonBind (GuiCharacter)
CharacterPlateGame:Bind (RulesetCharacter,TooltipDefinitions/AnchorMode)
CharacterPlateGameSelector:Bind (RulesetCharacter,bool,bool,bool,bool,CharacterPlateGameSelector/PlateSelectedHandler)
PartyControlPanel:Refresh ()
GuiPanel:RefreshNow ()
GuiPanel:Update ()
```

@RealBazou

ThyWoof:
image
@ChrisPJohn , after Bazou fixes were merge here. no more null object.

ThyWoof:
@PhilPJL : I checked this with other summonings and all cool. Can you check as well?

@RealBazou : are you ok with these changes?

PhilPJL (approved):

ChrisPJohn (approved):

PR #707, opened 2022-02-10
Suggested improvement/simplification

PhilPJL:
Putting this up for consideration/discussion:

1) ensure all future definitions have a unique name, then we can
2) add simplified constructors that create a definition from name only

### DIFF
--- a/SolastaCommunityExpansion/Api/BaseDefinitionBuilder.cs
+++ b/SolastaCommunityExpansion/Api/BaseDefinitionBuilder.cs
@@ -1,8 +1,10 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
+using HarmonyLib;
 using SolastaCommunityExpansion;
 using SolastaCommunityExpansion.Builders;
 using SolastaModApi.Diagnostics;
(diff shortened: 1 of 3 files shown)


ThyWoof (dismissed):

ThyWoof:
will let @ChrisPJohn review as well and merge.

SolastaCommunityExpansion/Api/BaseDefinitionBuilder.cs
```diff
@@ -31,6 +33,61 @@ protected static void LogDefinition(string msg)
...
+                return;
+            }
+
+            if (definitionNames.ContainsKey(name))
+            {
+                var definitionInDb = definitionNames[name];
+
+                var msg = definitionInDb ? "Non-CE definition:" : "CE definition";
+
+                throw new SolastaModApiException($"{msg}:{name} is already in use.");
```
ChrisPJohn, line 59: Should we throw the error in release builds or only debug?

On the one hand the error could help identify conflicts that cause issues, but it also could mean breaking CE because another mod introduces a name conflict that doesn't actually conflict (different dbs).

PhilPJL: I've improved the exception message.
Now throws in debug and logs in release.

ThyWoof (approved):

ChrisPJohn (approved):

PR #723, opened 2022-02-11
Multiclass in out logic

ThyWoof:
Improves MC in out logic calculation by ignoring FORCE attribute changes and subtracting any additional modifier added through wielded items.

### DIFF
--- a/SolastaCommunityExpansion/Multiclass/Models/InOutRulesContext.cs
+++ b/SolastaCommunityExpansion/Multiclass/Models/InOutRulesContext.cs
@@ -75,9 +75,14 @@ namespace SolastaCommunityExpansion.Multiclass.Models
             return UnityEngine.Mathf.Clamp(currentValue, minValue, attribute.MaxEditableValue > 0 ? attribute.MaxEditableValue : attribute.MaxValue);
         }
 
-        private static void EnumerateAttributeModifiers(RulesetCharacterHero hero, Dictionary<string, int> attributeModifiers)
+        private static void EnumerateItemsAttributeModifiers(RulesetCharacterHero hero, Dictionary<string, int> attributeModifiers)
         {
             var items = new List<RulesetItem>();
+
+            foreach (var attributeName in CoreAttributes)
+            {
+                attributeModifiers.Add(attributeName, 0);
+            }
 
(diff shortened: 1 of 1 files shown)


SolastaCommunityExpansion/Multiclass/Models/InOutRulesContext.cs
```diff
@@ -45,20 +46,74 @@ internal static void EnumerateHeroAllowedClassDefinitions(RulesetCharacterHero h
...
+            var items = new List<RulesetItem>();
+
+            hero.CharacterInventory.EnumerateAllItems(items, considerContainers: false);
+
+            foreach (var attributeName in CoreAttributes)
+            {
+                attributeModifiers.Add(attributeName, 0);
+            }
+
+            foreach (var featureDefinitionAttributeModifier in items
```
PhilPJL, line 90: You could use

```
            var coreAttributes = new HashSet<string>
            {
                AttributeDefinitions.Strength,
                AttributeDefinitions.Dexterity,
                AttributeDefinitions.Intelligence,
                AttributeDefinitions.Wisdom,
                AttributeDefinitions.Charisma
            };

            var dictionary = items
                    .SelectMany(item => item.ItemDefinition.StaticProperties.Select(sp => sp.FeatureDefinition))
                    .OfType<FeatureDefinitionAttributeModifier>()
                    .Where(fdam => coreAttributes.Contains(fdam.ModifiedAttribute))
                    .GroupBy(fdam => fdam.ModifiedAttribute)
                    .ToDictionary(g => g.Key, g => g.Sum(x => x.ModifierValue));
```
which would give you an inner join, whereas I think you want/have a left outer join (if I remember correctly)
What you have looks ok although I would go for the hashset.

PhilPJL (approved):

SolastaCommunityExpansion/Multiclass/Models/InOutRulesContext.cs
ThyWoof: I need a left join

ThyWoof: Which is the same as left outer join AFAIK

PhilPJL (dismissed):

PhilPJL (approved):

PR #726, opened 2022-02-12
Additional backstory display

ThyWoof:


ThyWoof:
image

SolastaCommunityExpansion/Patches/GameUi/CharacterInspection/GuiCharacterPatcher.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            if (additionalBackstory == null || additionalBackstory == string.Empty)
+            {
+                return;
+            }
+
+            var builder = new StringBuilder();
+
+            builder.Append(__result);
+            builder.Append("\n\n<B>ADDITIONAL BACKSTORY</B>\n\n");
```
ChrisPJohn, line 35: Additional Backstory should be in translations.

ChrisPJohn (changes requested):

SolastaCommunityExpansion/Patches/GameUi/CharacterInspection/GuiCharacterPatcher.cs
PhilPJL: Not sure but is it worth using Environment.Newline instead of \n since the game runs on Linux?  Probably no difference but...

PhilPJL (approved):

ChrisPJohn (approved):

PR #736, opened 2022-02-12
Add Configure and builder type to BaseDefinitionBuilder: option3

PhilPJL:
Compared to #734 #735 this doesn't require any type params when using .Configure, but does require updating 200+ references to BaseDefinitionBuilder to pass in the parent class type.
I think I prefer option 2.


PhilPJL:
Decided to go with option 2

PR #739, opened 2022-02-13
Beta standardize builder ctors

PhilPJL:
Continuing refactoring,
Added Configure method to base class which I used to refactor FeatureDefintionBuilder.
This is now hitting Chris's code more, so prefer Chris to approve (or not).

### DIFF
--- a/SolastaCommunityExpansion/Api/BaseDefinitionBuilder.cs
+++ b/SolastaCommunityExpansion/Api/BaseDefinitionBuilder.cs
@@ -502,6 +502,12 @@ namespace SolastaModApi
         {
             Assert.IsNotNull(configureDefinition);
             configureDefinition.Invoke(Definition);
+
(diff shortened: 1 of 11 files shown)


SolastaCommunityExpansion/Api/BaseDefinitionBuilder.cs
```diff
@@ -496,6 +496,20 @@ void VerifyGuiPresentation()
         #endregion
 
         protected TDefinition Definition { get; }
+
+        internal TBuilder Configure<TBuilder>(Action<TDefinition> configureDefinition)
```
ChrisPJohn, line 500: I feel like the style of using this to configure a definition is a different direction than creating builders for classes. Is this intended to be an eventual replacement for the builders  (eg: CharacterClassDefinitionBuilder) or along side them?

ChrisPJohn (approved):

PR #747, opened 2022-02-14
Don't use obsolete properties in extensions.

ChrisPJohn:


PhilPJL (changes requested):
Good idea but unfortunately the obsolete properties are still present (they've been recreated as SetChildrenStates, SetChildrenStateIndexes) and bypass the Obsolete warning by using SetField.

PhilPJL (approved):

PR #748, opened 2022-02-14
Tinkerer scout sentinel prep spells

ChrisPJohn:


SolastaCommunityExpansion/Classes/Tinkerer/Subclasses/ScoutSentinelSubClass.cs
```diff
@@ -16,9 +16,10 @@ public static class ScoutSentinelTinkererSubclassBuilder
         public const string Name = "ScoutSentinel";
         public const string Guid = "fb2e5f73-d552-430f-b329-1f0a2ecdf6bd";
 
+        // ArmorModePool needs to be instantiated before the two modes.
+        public static readonly FeatureDefinitionPower ArmorModePool = CreateArmorModePool();
```
ChrisPJohn, line 20: @PhilPJL FYI the change to making things readonly included changes to the execution order of code and broke this.

PhilPJL: Ok.  Ideally the dependencies between properties should create them in the right order.  I'll have to try to understand what went wrong in order to avoid in future.

PhilPJL (approved):

SolastaCommunityExpansion/Classes/Tinkerer/Subclasses/ScoutSentinelSubClass.cs
DubhHerder: Thanks for tagging me, I hadn't heard anything about changing the structure of the subclass like this. I would have been happy to look over it if I had. Changes like removing and restructuring/reordering of builders should really be tested or at least discussed to avoid in the future.

DubhHerder: do you know what commit the structure was changed so i can review?

DubhHerder: found it, https://github.com/ChrisPJohn/SolastaCommunityExpansion/pull/560/

PhilPJL: Thanks.  Did some testing.  It appears I was working under a bad assumption that static properties would get initialised in the order they were referenced, but they get initialised in the order they appear.

To make this work the properties have to be lazy loading as used in some places, e.g.

```
    private static FeatureDefinitionPower _armorModePool;
    public static FeatureDefinitionPower ArmorModePool => _armorModePool ??= CreateArmorModePool();
```

If all properties use this pattern then they only get created once and in the order of use which is ideal.  Shame there isn't a neat syntax like say 
```   
   public static FeatureDefinitionPower ArmorModePool => ??= CreateArmorModePool();
```

Looks like I have some more work to do.

PhilPJL: An alternative is to use Lazy<T> like this

`	public static readonly Lazy<FeatureDefinitionPower> ScoutModePower  = new Lazy<FeatureDefinitionPower>(CreateScoutModePower);
`

and then this is used as `ScoutModePower.Value`

That works as well.

PR #784, opened 2022-02-18
Beta: add some custom extension methods

PhilPJL:
Adding methods for SetRange, SetDuration. Do they make sense?


ThyWoof (approved):

ThyWoof (approved):

SolastaCommunityExpansion/Builders/EffectDescriptionBuilder.cs
```diff
@@ -269,6 +284,20 @@ public EffectDescriptionBuilder SetOffsetImpactTime(float offsetImpactTimeBasedO
             return this;
         }
 
+        public EffectDescriptionBuilder SetDuration(RuleDefinitions.DurationType type, int? duration = null)
+        {
```
ChrisPJohn, line 288: Did you see the SetDurationData and SetTargetingData (includes range) above?

SolastaCommunityExpansion/Api/AdditionalExtensions/EffectDescriptionExtensions.cs
```diff
@@ -0,0 +1,136 @@
+﻿using System;
+using SolastaModApi.Diagnostics;
+
+namespace SolastaModApi.Extensions
+{
+    public static partial class EffectDescriptionExtensions
+    {
+        public static T SetDuration<T>(this T entity, RuleDefinitions.DurationType type, int? duration = null)
```
ChrisPJohn, line 8: Do we want to support methods like this in class extensions or only in the builders? If builders become simply one to one passthroughs to methods added in extensions, the builders aren't adding very much benefit.

SolastaCommunityExpansion/Api/AdditionalExtensions/EffectDescriptionExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+                    }
+                    entity.SetDurationParameter(duration.Value);
+                    break;
+                default:
+                    if (duration != null)
+                    {
+                        throw new SolastaModApiException($"A duration value is not expected for duration type {type}");
+                    }
+                    // TODO: is this sensible?
+                    entity.SetDurationParameter(0);
```
ChrisPJohn, line 29: I'd go with the default value which (in the last decompile I have) is 1.

SolastaCommunityExpansion/Builders/ConditionDefinitionBuilder.cs
```diff
@@ -136,9 +136,13 @@ public TBuilder SetDuration(RuleDefinitions.DurationType type, int? duration = n
                     {
                         throw new SolastaModApiException($"A duration value is not expected for duration type {type}");
                     }
+                    // TODO: is this sensible?
+                    Definition.SetDurationParameter(0);
```
ChrisPJohn, line 140: I'd suggest the default value of 1.

SolastaCommunityExpansion/Builders/ConditionDefinitionBuilder.cs
```diff
@@ -115,10 +115,10 @@ public TBuilder ClearRecurrentEffectForms()
             return (TBuilder)this;
         }
 
+        // TODO: is it worth/possible to factor out ConditionDefinitionBuilder.SetDuration and EffectDescriptionBuilder.SetDuration?
```
ChrisPJohn, line 118: Static function that takes an object and checks what type the object is and/or if the appropriate fields exist before setting them?

Alternatively static function to do the validation of the values (it seems the validation here is the biggest chunk of code). If they validate then the builder sets the values on the object.

ChrisPJohn (changes requested):
Marking request changes for at least the default duration parameter.

SolastaCommunityExpansion/Builders/EffectDescriptionBuilder.cs
PhilPJL: No saw that later after raised the PR.

SolastaCommunityExpansion/Api/AdditionalExtensions/EffectDescriptionExtensions.cs
PhilPJL: ok I'll add a TODO to remove those extensions

SolastaCommunityExpansion/Builders/ConditionDefinitionBuilder.cs
PhilPJL: I don't think the default values are meaningful, e.g.
image
I think I'll stick with 0 for these, although the values will be ignored in these cases (if I've got it right).

SolastaCommunityExpansion/Builders/ConditionDefinitionBuilder.cs
PhilPJL: I like the validation idea.

SolastaCommunityExpansion/Builders/EffectDescriptionBuilder.cs
PhilPJL: Will combine in future

SolastaCommunityExpansion/Api/AdditionalExtensions/EffectDescriptionExtensions.cs
PhilPJL: Need to fix various places to use a builder instead of the definition directly

SolastaCommunityExpansion/Api/AdditionalExtensions/EffectDescriptionExtensions.cs
PhilPJL: Sorry sticking with 0 for values which should have no duration/range.

SolastaCommunityExpansion/Api/AdditionalExtensions/EffectDescriptionExtensions.cs
ChrisPJohn: Either path is valid. I was just raising this because I thought one of our goals was to be moving towards more consistency with a single suggested way to do things.

ChrisPJohn (approved):

PR #785, opened 2022-02-18
Beta: implement square cylinder

PhilPJL:
Implement a 'Square Cylinder' TargetType by patching code using TargetType.Cube.
If TargetParameter2 != 0 then adjust geometry code to use a square cylinder of height TargetParameter2.

TODO: Adjust wire frame drawn showing area of effect after casting the spell - help needed.
TODO: I assume transpilers will reduce the number of patches. Wait until logic fully tested.


ThyWoof:
I believe all patterns Prefix / Postfix that set Height can be done with less patches if we transpile the call to `UpdateCubePosition_Regular` within `UpdateGeometricShape`.

SolastaCommunityExpansion/Settings.cs
```diff
@@ -169,6 +169,10 @@ public class Settings : UnityModManager.ModSettings
         public bool DismissControlledConjurationsWhenDeliberatelyDropConcentration { get; set; }
         public bool OnlyShowMostPowerfulUpcastConjuredElementalOrFey { get; set; }
         public bool SpikeGrowthDoesNotAffectFlyingCreatures { get; set; }
+        public bool SquareAreaOfEffectSpellsDoNotAffectFlyingCreatures { get; set; } = true;
```
ChrisPJohn, line 172: This should probably not be set to true by default.

SolastaCommunityExpansion/Spells/HouseSpellTweaks.cs
```diff
@@ -28,17 +29,101 @@ private static void SpikeGrowthDoesNotAffectFlyingCreatures()
...
+                    .GetDatabase<SpellDefinition>()
+                    .Where(sd => sd.EffectDescription.TargetType == RuleDefinitions.TargetType.Cube))
+                {
+                    // TargetParameter2 is not used by TargetType.Cube but has random values assigned.
+                    // We are going to use it to create a square cylinder with height so set to zero for all spells with TargetType.Cube.
+                    sd.EffectDescription.SetTargetParameter2(0);
+                }
+            }
+
+/*            static void LogAllSpellsWithTargetTypeCube()
```
ChrisPJohn, line 111: Clean up debug code?

ChrisPJohn (changes requested):
A couple small changes.

For GeometryPatcher I focused on if the code was protected with settings checks and less on the math. I assume I'm unlikely to catch a math error that wasn't caught testing the changes, but also it's late. If you want a thorough math review I can take a look later.

SolastaCommunityExpansion/Spells/HouseSpellTweaks.cs
PhilPJL: Removed

ChrisPJohn (approved):

ThyWoof (approved):

Issue #717, opened 2022-02-11
From Nexus: Minor Lifesteal Does Not Provide Temp HP


ChrisPJohn:
flerpidoodle
The description of minor lifesteal says it grants the stolen health as temp HP to the caster; in reality, it heals the caster. This might just be a documentation bug.

mattelkinsTC:
Is the desired behavior to heal or to grant temp HP? I'm considering dipping a toe into helping out, if I can find the time. Not sure whether to patch the spell description or the spell.

ChrisPJohn:
I think the original idea was to have it grant temp HP, but there were implementation issues with that. @RealBazou might have more information.

PR #790, opened 2022-02-19
Issue #717: Minor Lifesteal Does Not Provide Temp HP

mattelkinsTC:
After looking into this a bit, I decided that fixing the spell to performing like the description would be biting off more than I wanted to chew for a first contribution, especially since I am not yet familiar with Unity. The challenge is in the fact that there are no rule definitions for the desired behavior. Instead, I opted for the simple "fix" of changing the description to match the actual behavior; at least this way it isn't misleading.

Attached are before and after screenshots.

lifesteal-before
lifesteal-after


ChrisPJohn:
That looks fine with me. I added Bazou as the official reviewer though since it is his spell implementation.

RealBazou (approved):

PR #792, opened 2022-02-19
Beta refactor

PhilPJL:
All builders now use correct base classes.


SolastaCommunityExpansion/Builders/Features/FeatureDefinitionAdditionalDamageBuilder.cs
```diff
@@ -8,144 +8,187 @@
...
+            Definition.SetNotificationTag(notificationTag);
+            Definition.SetLimitedUsage(limitedUsage);
+            Definition.SetDamageValueDetermination(damageValueDetermination);
+            Definition.SetTriggerCondition(triggerCondition);
+            Definition.SetRequiredProperty(requiredProperty);
+            Definition.SetAttackModeOnly(attackModeOnly);
+
+            SetDamageDice(damageDieType, damageDiceNumber);
+
+            // Does this conflict with SetSpecificDamageType below?
```
ChrisPJohn, line 147: No it doesn't conflict. Additional damage type is an enum, with one of the options being "Specific". If Specific is the additional damage type then the game reads the specific damage type.

SolastaCommunityExpansion/Level20/Features/ProficiencyRogueBlindSenseBuilder.cs
```diff
@@ -1,10 +1,10 @@
-﻿using SolastaCommunityExpansion.Builders;
+﻿using SolastaCommunityExpansion.Builders.Features;
 using SolastaModApi.Extensions;
 using static SolastaModApi.DatabaseHelper.FeatureDefinitionSenses;
 
 namespace SolastaCommunityExpansion.Level20.Features
 {
-    internal sealed class ProficiencyRogueBlindSenseBuilder : DefinitionBuilder<FeatureDefinitionSense>
+    internal sealed class ProficiencyRogueBlindSenseBuilder : FeatureDefinitionBuilder<FeatureDefinitionSense, ProficiencyRogueBlindSenseBuilder>
```
DubhHerder, line 7: Not sure if this is worth noting or not.: this change doesnt seem to match the pattern established for previous changes to base game feature builders. It looks closer to the pattern for custom features but is not a custom feature. 
Edit RangerFeralSensesBuilder also breaks the pattern so maybe it's normal and related to how FeatureDefinitionSense are currently being built.

DubhHerder (dismissed):

SolastaCommunityExpansion/Level20/Features/ProficiencyRogueBlindSenseBuilder.cs
PhilPJL: The builders aren't all consistent yet, so they don't all have a pair of classes, and they don't have to.  I may get round to it though.  You've identified a builder that isn't in the Builders/Features folder, so if we want we can add

// this one is to create builders for definitions derived from FeatureDefinitionSense 
abstract class FeatureDefinitionSenseBuilder ; FeatureDefinitionBuilder<TDefinition, TBuilder>
where TDefinition : FeatureDefinitionSense (which is a FeatureDefinition)
{
...
}

// this one is a builder to create/copy FeatureDefinitionSense 
class FeatureDefinitionSenseBuilder : 
    FeatureDefinitionSenseBuilder<FeatureDefinitionSense,FeatureDefinitionSenseBuilder>
{
...

Create...
}

// we could then have a specialized instance, e.g.
class ProficiencyRogueBlindSenseBuilder : FeatureDefinitionSenseBuilder {}

// But since this doesn't actually do anything special we could instead use
ProficiencyRogueBlindSense = FeatureDefinitionSenseBuilder.Create(SenseBlindSight2, name, guid).SetGuiPresentation(...).AddToDB()

So yes, more work.

DubhHerder: I assumed that was the case and I dont think it is an issue. Just wanted to confirm.

ThyWoof (approved):

PR #797, opened 2022-02-19
transpile patches

ThyWoof:
@PhilPJL : I'm assuming in both cases that height won't be null so changed datatype in both scenarios. please confirm.


PhilPJL (dismissed):
Looks good so far.  Will double check other spells later.

PhilPJL (approved):

PR #798, opened 2022-02-20
Change non-variable duration to 0.

PhilPJL:
Change the duration value of 1 used for non-variable duration types to 0.

### DIFF
--- a/SolastaCommunityExpansion/Builders/EffectDescriptionBuilder.cs
+++ b/SolastaCommunityExpansion/Builders/EffectDescriptionBuilder.cs
@@ -103,34 +103,34 @@ namespace SolastaCommunityExpansion.Builders
         }
 
         // TODO: combine with SetTargetingData - this method not currently used, the extension version is used
-/*        private EffectDescriptionBuilder SetRange(RuleDefinitions.RangeType type, int? range = null)
-        {
-            switch (type)
-            {
-                case RuleDefinitions.RangeType.RangeHit:
-                case RuleDefinitions.RangeType.Distance:
-                    if (range == null)
+        /*        private EffectDescriptionBuilder SetRange(RuleDefinitions.RangeType type, int? range = null)
+                {
+                    switch (type)
                     {
-                        throw new ArgumentNullException(nameof(range), $"A range value is required for range type {type}.");
+                        case RuleDefinitions.RangeType.RangeHit:
+                        case RuleDefinitions.RangeType.Distance:
+                            if (range == null)
+                            {
+                                throw new ArgumentNullException(nameof(range), $"A range value is required for range type {type}.");
+                            }
+                            effect.SetRangeParameter(range.Value);
+                            break;
+                        case RuleDefinitions.RangeType.Touch:
+                            effect.SetRangeParameter(range ?? 0);
+                            break;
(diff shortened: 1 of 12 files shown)


SolastaCommunityExpansion/Builders/EffectDescriptionBuilder.cs
```diff
@@ -103,34 +103,34 @@ public EffectDescriptionBuilder SetTargetingData(RuleDefinitions.Side targetSide
...
+                            }
+                            effect.SetRangeParameter(range.Value);
+                            break;
+                        case RuleDefinitions.RangeType.Touch:
+                            effect.SetRangeParameter(range ?? 0);
+                            break;
+                        default: // Self, MeleeHit
+                            if (range != null)
+                            {
+                                throw new SolastaModApiException($"A duration value is not expected for duration type {type}");
```
DubhHerder, line 124: should this refer to range instead of duration?

DubhHerder (changes requested):
For the individually created effect descriptions, changing from 1 to 0 for this parameter seems like less of a fix and more of a personal preference, and if that's the case my personal preference (at least for the code i added) is to not make unnecessary changes and keep it to 1. Has each effect been tested after being changed to ensure there's no change in gameplay? 
If not i'd prefer to keep code that has been tested in the game as some of my code pushes at the boundaries of the base games features and uses them in unintended ways to minimise the number of patches required.

SolastaCommunityExpansion/Builders/EffectDescriptionBuilder.cs
PhilPJL: yes copy/paste error

PhilPJL:
> For the individually created effect descriptions, changing from 1 to 0 for this parameter seems like less of a fix and more of a personal preference, and if that's the case my personal preference (at least for the code i added) is to not make unnecessary changes and keep it to 1. Has each effect been tested after being changed to ensure there's no change in gameplay? If not i'd prefer to keep code that has been tested in the game as some of my code pushes at the boundaries of the base games features and uses them in unintended ways to minimise the number of patches required.

In the game code the duration value isn't used unless it's one of the four variable duration types (Round, Minute, Hour, Day). I can change my assertion to ignore whatever duration value is used or remove it altogether if preferred.  The '1' has crept in because of the default values on EffectDescription which are inconsistent, e.g.
image

DubhHerder:
> In the game code the duration value isn't used unless it's one of the four variable duration types (Round, Minute, Hour, Day). I can change my assertion to ignore whatever duration value is used or remove it altogether if preferred. The '1' has crept in because of the default values on EffectDescription which are inconsistent, e.g.

Yes I know the duration value isn't used unless it's one of those types, that's one reason why I see no need to change it. 
The other reason is that I'm not a fan of making untested changes to features that have been previously play-tested. Most of the time it's probably fine but now and then unintended effects will occur if it's done often enough . Makes me less confident that the feature/spell etc works as I intend and increases the chances that it'll take more time to find and fix any bugs that do get reported if i am no longer familiar with the related code. 
Also I'd like to catch up to all previous changes that have occurred to the code that people likely expect me to fix if there is a bug before further changes are made. I missed many of the previous changes that werent as automated as i first thought they would be.

PR #801, opened 2022-02-20
Bring back handing the fighting style feat addition.

ThyWoof:


SolastaCommunityExpansion/Models/CustomFeaturesContext.cs
```diff
@@ -19,6 +19,26 @@ internal static void RecursiveGrantCustomFeatures(RulesetCharacterHero hero, Lis
...
+                if (!(grantedFeature is FeatureDefinitionProficiency featureDefinitionProficiency))
+                {
+                    continue;
+                }
+
+                if (featureDefinitionProficiency.ProficiencyType != RuleDefinitions.ProficiencyType.FightingStyle)
+                {
+                    continue;
+                }
+                using (var enumerator = featureDefinitionProficiency.Proficiencies.GetEnumerator())
```
PhilPJL, line 32: This should probably be

```
foreach(var proficiency in featureDefinitionProficiency.Proficiencies)
{
    var definition = DatabaseRepository.GetDatabase<FightingStyleDefinition>().GetElement(key, false);

    if(definition == null)
    {
        throw new SolastaModApiException("... ");
    } 
hero.TrainedFightingStyles.Add(definition);
```

    
}

PhilPJL: GetElement does this which really isn't very helpful if the key doesn't exist

```
    public T GetElement(string key, bool allowFail = false)
    {
        T value = null;
        if (valuesTable.TryGetValue(key, out value))
        {
            return value;
        }

        if (!allowFail)
        {
            Trace.LogError("Cannot find element {0} in database {1}", key, Name);
        }

        return value;
    }
```

PhilPJL (changes requested):
See comment

SolastaCommunityExpansion/Models/CustomFeaturesContext.cs
ChrisPJohn: I agree a for loop would be better. I was simply bringing back code that was wrongly deleted in a refactor.

ChrisPJohn:
I didn't make this pull request I'm closing it.

PR #805, opened 2022-02-21
Beta diagnostics

PhilPJL:
Reworked current Gui presentation and definition diagnostics, and added serialization of all CE definitions to json.

Set the system environment variable 'SolastaCEDiagnosticsDir' to point to you project diagnostics folder, e.g.
"C:\Users\passp\source\repos\SolastaCommunityExpansion\SolastaCommunityExpansion\Diagnostics"

When you run the game all gui presentations and definitions will be saved in 6 files
image

If these files have changes you will see them before commit and can then check nothing's fup'ed.
Please give it a try. Check one of your definitions and confirm it's serialized correctly in CE-Definitions.json.  Check that if you screw something it shows up.


SolastaCommunityExpansion/Settings.cs
```diff
@@ -232,23 +232,15 @@ public class Settings : UnityModManager.ModSettings
         public bool NoExperienceOnLevelUp { get; set; }
 
         // Debug
-        public bool DebugShowTADefinitionsWithMissingGuiPresentation { get; set; }
         public bool DebugLogDefinitionCreation { get; set; }
         public bool DebugLogFieldInitialization { get; set; }
         public bool DebugDisableVerifyDefinitionNameIsNotInUse { get; set; }
-        public List<string> KnownDuplicateDefinitionNames { get; } = new()
+        internal List<string> KnownDuplicateDefinitionNames { get; } = new()
```
ThyWoof, line 238: If you make this internal the setting won’t dump to the xml file. It might be useful to tweak the behavior without reconciling. If not the intent then might be better to move this into Diagnostics.

ThyWoof (changes requested):

SolastaCommunityExpansion/Settings.cs
PhilPJL: Yes I know, every time the settings were saved that value was being added.  I had a list of about 50 "SummonProtectorConstruct" in my settings file. Couldn't see how to stop that behaviour.  I can move it for now I guess.

DubhHerder (approved):
I am not familiar with this type of process so i cant offer much in the way of reviewing other than noting that TA are obviously comfortable having missing translations / guiPresentations so is there a need to track their omissions or us?

PhilPJL:
> I am not familiar with this type of process so i cant offer much in the way of reviewing other than noting that TA are obviously comfortable having missing translations / guiPresentations so is there a need to track their omissions or us?

Not sure tracking TA's missing translations etc is of any benefit other than we can tell them, although surely they know already? But I quite often see 'Feature/&xyz' in the game so they do need to fix them.

SolastaCommunityExpansion/Settings.cs
PhilPJL: Done

ThyWoof (dismissed):

SolastaCommunityExpansion/Settings.cs
```diff
@@ -232,23 +232,11 @@ public class Settings : UnityModManager.ModSettings
...
-#if DEBUG
-        public bool DebugShowCEDefinitionsWithMissingGuiPresentation { get; set; } = true;
-        public bool DebugLogCEDefinitionsToFile { get; set; } = true;
-        public bool DebugLogCEGuiPresentationsToFile { get; set; } = true;
-#else
-        internal bool DebugShowCEDefinitionsWithMissingGuiPresentation { get; set; }
-        internal bool DebugLogCEDefinitionsToFile { get; set; }
-        internal bool DebugLogCEGuiPresentationsToFile { get; set; }
-#endif
+        public bool DebugEnableTADefinitionDiagnostics { get; set; } = true;
```
ChrisPJohn, line 238: Should these just be true for debug and false for release?

SolastaCommunityExpansion/Diagnostics/CE-Definitions-GuiPresentation-MissingTranslation-en.txt
```diff
@@ -0,0 +1,95 @@
+AddConstructCantrips	Title='Feat/&AddConstructCantripsTitle'.
```
ChrisPJohn, line 1: I assume we can use this to track what keys need translations when/if we start supporting localizing CE?

Are we expecting to work on making the en list 0 items or is it ok for stuff to be in here?

ChrisPJohn (changes requested):
Main thing is defaulting this to off for release builds since I assume we don't need this in releases.

SolastaCommunityExpansion/Settings.cs
PhilPJL: I'm going to change those to false and add UI.

SolastaCommunityExpansion/Diagnostics/CE-Definitions-GuiPresentation-MissingTranslation-en.txt
PhilPJL: This is a hint for use to add translations or change the GUIPresentation to indicate no translation required.

SolastaCommunityExpansion/Settings.cs
PhilPJL: Not sure where to put debug/diagnostics settings @ThyWoof

SolastaCommunityExpansion/Diagnostics/CE-Definitions-GuiPresentation-MissingTranslation-en.txt
PhilPJL: Some of them will drop out after #810 
Some of them are erroneous because I notice they are explicitly and empty string in translations.txt.  
Need to remove Telema ones
Some like PowerAttackCondition don't have a key, just text - odd?
But the rest should have English entries or should explicitly be Feature/NoContent

PhilPJL: I see that TA have "Feature/&Emptystring".  After this and #810 are in, I'll add that and take it into account in diagnostics.

ChrisPJohn (approved):

SolastaCommunityExpansion/Diagnostics/CE-Definitions-GuiPresentation-MissingTranslation-en.txt
```diff
@@ -0,0 +1,70 @@
...
+ScoutSentinelFeatureSet_level15	Title='Feat/&ScoutSentinelFeatureSet_level15Title'.
+SelfDestructionCondition	Title='Rules/&SelfDestructionConditionTitle'.
+SelfDestructionCondition	Description='Rules/&SelfDestructionConditionDescription'.
+SummonProtectorConstruct_Upgrade	Title='Feat/&SummonProtectorConstructTitle_Upgrade'.
+SummonProtectorConstruct_Upgrade	Description='Feat/&SummonProtectorConstructDescription_Upgrade'.
+SummonProtectorPowerConstruct_Upgrade	Title='Feat/&SummonProtectorConstructTitle_2'.
+SummonProtectorPowerConstruct_Upgrade	Description='Feat/&SummonProtectorConstructDescription_2'.
+TelemaDemoUnleashed	Title='Telema Unleashed'.
+TelemaDemoUnleashed	Description='This mini campaign allows you to play the Kickstarter Telema demo.'.
+TraditionSpellMasterSpellResistance	Title='Spell/&TraditionSpellMasterSpellResistanceTitle'.
```
ChrisPJohn, line 58: Just an FYI.

So this looks like a refactoring broke it at some point. TraditionSpellMasterSpellResistanceTitle and description are under Subclass in the translations file (not spell).

I don't know if other entries in here are similar.

PhilPJL: I'll check

PR #807, opened 2022-02-22
minor tweaks

ThyWoof:


ThyWoof:
* use AttributeDefinitions instead of hard-coded string
* protect 3 multiclass methods with global setting
* don't display translation warnings on release build

ChrisPJohn (approved):
Approval for the updates to custom features.

PhilPJL (approved):

PR #811, opened 2022-02-22
Add empty string

PhilPJL:
Set explicit nocontent on Dummy class and add GuiPresentationBuilder.EmptyString to be used to indicate empty string is being used explicitly.


DubhHerder (commented):
As it involves multiclass, should @ThyWoof be added as a reviewer?

ThyWoof (approved):

ThyWoof:
This is only a sentinel to simplify MC checks. Ok to add gui presentation there.

PR #813, opened 2022-02-22
fix twf

whiplashomega:
Fixes the Two Weapon Fighting style feat to correctly add ability modifier damage to off-hand weapon attacks by importing the core two weapon fighting style modifier feature into the feat.


ChrisPJohn (changes requested):
So I understand this works for two weapon fighting and because two weapon fighting only provides a benefit when the fighting style is valid (at least I think so), this shouldn't have unintended side effects. However the core issue is this change:
https://github.com/ChrisPJohn/SolastaCommunityExpansion/pull/533/files
(specifically removing this code: https://github.com/ChrisPJohn/SolastaCommunityExpansion/blob/434d43c9a19bec792caeb4e58ea861639542a4d9/SolastaCommunityExpansion/Patches/FightingStyleFeats/RulesetCharacterHeroPatcher.cs )
broke how all of the fighting style feats work and the code that was accidentally deleted needs to be brought back.

PR #815, opened 2022-02-24
update Dungeon Maker placement features

ThyWoof:
Tweaked approach from changing a few DM blueprints to allow anything to be placed anywhere when CTRL is pressed.

ps: we don’t need this on dev as it isn’t a bug fix.

### DIFF
--- a/SolastaCommunityExpansion/Models/DungeonMakerContext.cs
+++ b/SolastaCommunityExpansion/Models/DungeonMakerContext.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.IO;
 using System.Linq;
-using SolastaModApi.Extensions;
 
 namespace SolastaCommunityExpansion.Models
 {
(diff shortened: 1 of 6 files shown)


ChrisPJohn:
> Tweaked approach from changing a few DM blueprints to allow anything to be placed anywhere when CTRL is pressed.
> 
> ps: we don’t need this on dev as it isn’t a bug fix.

I thought you told folks this would be in the next CE release. My understanding was we were going to do a dev release for some bug fixes (and other changes). Should this be in dev?

ThyWoof:
> > Tweaked approach from changing a few DM blueprints to allow anything to be placed anywhere when CTRL is pressed.
> > ps: we don’t need this on dev as it isn’t a bug fix.
> 
> I thought you told folks this would be in the next CE release. My understanding was we were going to do a dev release for some bug fixes (and other changes). Should this be in dev?

Ok for one more dev release but I don’t have the time to test this change in dev. Many authors are waiting anyways for the DLC.

ChrisPJohn (approved):

PhilPJL (approved):

PR #821, opened 2022-02-24
Beta diagnostics

PhilPJL:
Spacey's serialization code converts referenced definitions to references like
"requiredTargetCondition": "Definition:ConditionMarkedByArcanist:2c45ce76-761a-5335-83ce-f380422be675"
and also prevents recursion.  Seems to work fine when serializing definitions in isolation but not when serializing an array.

As a quick fix I'm serializing each definition individually and have also turned of PreserveReferencesHandling.  This takes longer and results in a larger file, but is hopefully guaranteed to catch everything.

I would like to add a button to explicitly run this export which takes 15-20 seconds.

Merged DataMiner.  4 buttons in help&credits viewer to create blueprints and diagnostics.


ThyWoof:
We should not merge Dataminer as we need it standalone to only export official content.

unless you fixed that…

PhilPJL:
> We should not merge Dataminer as we need it standalone to only export official content.
> 
> unless you fixed that…

Yes assuming only CE is running we capture the TA definitions before any of ours get loaded.

ThyWoof:
Perfect. Assuming there will only be CE and Holic’s work I believe it’s a safe assumption as they will load after us

ThyWoof:
We can add DM button to display in debug mode only next to Unity Explorer button on Help UI

PhilPJL:
> Perfect. Assuming there will only be CE and Holic’s work I believe it’s a safe assumption as they will load after us

Even with DM we normally have to unload all other mods to be sure anyway.

SolastaCommunityExpansion/Models/DiagnosticsContext.cs
```diff
@@ -146,7 +149,21 @@ public static void PostCELoad()
                 ceDefinitions.Select(d => $"{d.Name}, {d.GUID}"));
 
             // Write all CE definitions to file (json)
-            ExportDefinitions(ceDefinitions, Path.Combine(DiagnosticsOutputFolder, "CE-Definitions.json"));
+            ExportDefinitions(Path.Combine(DiagnosticsOutputFolder, "CE-Definitions.json"), ceDefinitions);
+
+            /*
```
ChrisPJohn, line 154: Clean up commented code?

ChrisPJohn (commented):
We could make this only run manually (probably a button in mod settings?) if it's slowing down game start up a lot. It's not needed on every game start, though auto export means you can't forget to do it.

ThyWoof (approved):

PR #844, opened 2022-02-26
Beta diagnostics

PhilPJL:
Fix issue with blueprint dump.  (Json serializer needs to be recreated for each blueprint).
Add log for EffectForm verification.
Add ItemDescription verification and log.

Change SwitchMagicStaffFoci - not sure if this is correct since 3 items don't get Arcane.


ThyWoof:
That wasn't the correct behavior on foci items. Will add requires restart to the setting UI.

PR #847, opened 2022-02-27
Diagnostics Export alternate

ThyWoof:
. consolidated 2 exporters into one singleton
. CE now exports into individual files
. pressing button again cancels export
. moved from Help & Credits into Diagnostics & Patches
. fixed 2 null patchers in DEBUG mode
. defaults output to game folder if env var isn't set

image


ThyWoof:
@ChrisPJohn , changed nexus description dump to a file under diagnostics instead of mod log.

PhilPJL (approved):

PR #859, opened 2022-02-27
add more items to filter

ThyWoof:
@PhilPJL , I moved filter into cache phase. I don’t believe this is an issue as we don’t touch any of those blueprints. Also, do we still need to create TA-Definitions.txt / CE-Definitions.txt now that we have Assets.txt and Types.txt on corresponding subfolders?


ThyWoof:
I believe I know why. I had a settings.xml that was already initialized with the array data. Keep this in mind when you test. Need to remove entries from there.

PhilPJL (approved):
Works for me

PR #878, opened 2022-03-01
Elemental adept feat

ThyWoof:
Add Elemental Adept feats.


ThyWoof:
Closing this one as I would rather start using the delegate patches to solve logic here

PR #880, opened 2022-03-01
add welcome message on first start

ThyWoof:


ThyWoof:
@PhilPJL , EN isn’t my first language. Any ideas on a simple message to guide new users to open the mod UI?

PhilPJL:
> @PhilPJL , EN isn’t my first language. Any ideas on a simple message to guide new users to open the mod UI? Once we have it I’ll add the text to translations

First stab:

Welcome to Solasta + the Community Expansion Mod", "The CE mod is packed full of exciting features, but by default nothing is selected.  Press Ctrl-F10 to open the Mod Manager UI then click on the settings icon next to 'Community Expansion' to open the settings pages. Explore the options and select those you like.
image

ThyWoof:
We need to make that title smaller too. “Welcome to CE Mod”?

ThyWoof:
@PhilPJL , text added to Translations-en.txt. Also changed the logic to always display the msg under DEBUG mode (was using that for test) as well as display the message whenever a new mod version is installed.

We could later use this logic to present a change log.

SolastaCommunityExpansion/Patches/GameManagerPatcher.cs
```diff
@@ -94,7 +98,32 @@ internal static void Postfix()
...
+            var assemblyVersion = Assembly.GetExecutingAssembly().GetName().Version.ToString();
+
+#if RELEASE
+            if (Main.Settings.CurrentModVersion == assemblyVersion)
+            {
+                return;
+            }
+#endif
+
+            Main.Settings.CurrentModVersion = assemblyVersion;
```
ChrisPJohn, line 117: If we're going to force this on users on each patch we should make it useful for them (change log or even better- list of changes from their last update).

SolastaCommunityExpansion/Patches/GameManagerPatcher.cs
```diff
@@ -94,7 +98,32 @@ internal static void Postfix()
...
+
+            Gui.GuiService.ShowMessage(
+                MessageModal.Severity.Informative1,
+                "Message/&MessageModWelcomeTitle",
+                "Message/&MessageModWelcomeDescription",
+                "Message/&MessageOkTitle",
+                string.Empty,
+                () => UnityModManager.UI.Instance.ToggleWindow(),
```
ChrisPJohn, line 125: Can we have it open the window if they click a button to go to the settings and not force it on them? Also is it possible the window is already open and this call to ToggleWindow closes it?

SolastaCommunityExpansion/Patches/GameManagerPatcher.cs
ThyWoof: Changed the message to only display when there is no settings.xml file in the mod folder. We don't have much real state on this pop up message and the other option we can use has a timeout.

SolastaCommunityExpansion/Patches/GameManagerPatcher.cs
ThyWoof: Opening the Mod UI shouldn't be an issue now that we only display this message when there is no settings.xml file. Mod UI window starts open by default. When open the game cannot take any input as we patched that so it's guaranteed it won't be open.

ChrisPJohn (approved):

Issue #590, opened 2022-01-29
User report from Nexus: Double ability bonus for pushing


ChrisPJohn:
Dima793

The versions of Solasta, the Solasta Mod API, and Solasta Community Expansion.

﻿Solasta 1.2.15 - Public, Solasta Mod API 1.2.15.6, Solasta Community Expansion 1.2.15.7.

A list of other mods you have installed

SOLASTA DM Pro - Multiplayer 1.2.15.3, Divine Casting with Weapon and Shield 0.0.2.
A short description of the bug

Sorr-Akkath Saboteur has double the ability bonus for athletics when pushing. Extra bonus is reflected in the log as "+5: Sorr-Akkath Saboteur has: None".

A step-by-step procedure to reproduce it

Make player's character stand on an edge, let the enemy push it off a cliff. Conditions in which I experienced this bug:
Location: Caer Lem Outpost, after the rotating bridge, last fight of the "escape the caves" quest. Difficulty: Cataclysm, unmodified. The monster was fully lit, in case it's supposed to affect anything. Other monsters had proper modifiers.

The save, character and log files

Google drive folder: https://drive.google.com/drive/folders/1gFfwfI0sAQXlhEPeaUM-VZVjFfiYgWFc﻿.

DubhHerder:
The saboteur has proficiency in athletics in the monster definition:

```skillScores": [
    {
      "$id": "3",
      "$type": "MonsterSkillProficiency, Assembly-CSharp",
      "skillName": "Athletics",
      "bonus": 5
    }```

And a feature definition `AbilityCheckAffinitySaboteurStrengthBonus` which grants

```
"$type": "FeatureDefinitionAbilityCheckAffinity, Assembly-CSharp",
  "affinityGroups": [
    {
      "$id": "1",
      "$type": "FeatureDefinitionAbilityCheckAffinity+AbilityCheckAffinityGroup, Assembly-CSharp",
      "abilityScoreName": "Strength",
      "proficiencyName": "Athletics",
      "abilityCheckModifierDiceNumber": 5
    }
  ],```

The feature uses "Feature/&NoContentTitle" for it's title and description which is where the "has : none" in the combat log is coming from. The bug should probably be reported in the official bug channel first as it's not caused by mods, could be reopened if the Devs ignore it after the report

ThyWoof:
not a mod bug

PR #884, opened 2022-03-03
refactor indomitable might / reorganize patches.customFeatures folder

ThyWoof:
* Indomitable Might should change the check result to be equals to the strength attribute current value;
* Code was very specific and only considering this particular case. Changed interface to allow more flexibility for other features;
* There was also a bug report on Nexus about shoves using double bonus and this code was touching `AbilityCheck`;
* same number of patches after refactor (only split files and moved into sub folders);


DubhHerder:
The double bonus shove bug is limited to a specific monster and is due to the monsters features rather than any issue with the code, as described in issue #590

ThyWoof:
Thanks @DubhHerder. Anyway this code is much cleaner. It now supports any other attribute

SolastaCommunityExpansion/Patches/CustomFeatures/RulesetCharacterPatcher.cs
```diff
@@ -47,103 +55,136 @@ internal static void Postfix(RulesetCharacter __instance)
...
-            ref int minRoll)
+            List<RuleDefinitions.TrendInfo> advantageTrends,
+            List<RuleDefinitions.TrendInfo> modifierTrends,
+            ref int result)
         {
-            if (abilityScoreName != AttributeDefinitions.Strength)
-            {
-                return;
-            }
+            var featuresToBrowse = new List<FeatureDefinition>();
```
PhilPJL, line 75: could be

```
            var min = rulesetCharacter
                .EnumerateFeaturesToBrowse<IChangeAbilityCheck>()
                .Max(x => x.MinAbilityCheck(rulesetCharacter, baseBonus, rollModifier, abilityScoreName, proficiencyName, advantageTrends, modifierTrends));

            if (result < min)
            {
                result = min;
            }
```

I'll get my extension method used somehow :)

PhilPJL (commented):
It all looks ok, but I'll leave this to Chris.

SolastaCommunityExpansion/Patches/CustomFeatures/RulesetCharacterPatcher.cs
ThyWoof: Thanks. I didn't like the OfType used again.

ThyWoof:
@ChrisPJohn , once we get this one checked my next step is to reorganize the custom features folder to simplify matching new features with patches.

SolastaCommunityExpansion/Patches/CustomFeatures/RulesetCharacterPatcher.cs
PhilPJL: You can use the extension method as
`EnumerateFeaturesToBrowse<IChangeAbilityCheck>(true)`
to update the rulesetCharacter featuresToBrowse property.  Which you don't want in this case I think.

SolastaCommunityExpansion/Patches/CustomFeatures/ChangeAbilityCheck/RulesetCharacterPatcher.cs
```diff
@@ -0,0 +1,131 @@
...
+using SolastaModApi.Extensions;
+
+namespace SolastaCommunityExpansion.Patches.CustomFeatures.ChangeAbilityCheck
+{
+    //
+    // IChangeAbilityCheck
+    //
+    [HarmonyPatch(typeof(RulesetCharacter), "RollAbilityCheck")]
+    [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
+    internal static class RulesetCharacter_RollAbilityCheck
```
ChrisPJohn, line 17: Did you consider using a Postfix patch on RollDie to modify the die roll (in this case set to min if it's too low) and capturing context (really just abilityScore is needed for this use case) in prefixes on RollAbilityCheck and ResolveContestCheck?

That would allow for a more general die roll modifier if that's the goal of this change. It would require an additional patch for this feature, but no transpiler.  That approach would also make it possible to implement something like a Clockwork Amulet (modifies attack rolls) with no additional patches.

ChrisPJohn (changes requested):
Request changes for the CustomCode folder name and to consider the changes to the die roll patches.

SolastaCommunityExpansion/Patches/CustomFeatures/CustomCode/RulesetCharacterHeroPatcher.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using HarmonyLib;
+using SolastaCommunityExpansion.Models;
+
+namespace SolastaCommunityExpansion.Patches.CustomFeatures.CustomCode
+{
+    [HarmonyPatch(typeof(RulesetCharacterHero), "TrainFeats")]
+    [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
+    internal static class RulesetCharacterHero_TrainFeats
```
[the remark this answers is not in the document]
ThyWoof: Moved this patch back to top folder and added more comments there.

SolastaCommunityExpansion/Patches/CustomFeatures/CustomCode/CharacterBuildingManagerPatcher.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using HarmonyLib;
+
+namespace SolastaCommunityExpansion.Patches.CustomFeatures.CustomCode
```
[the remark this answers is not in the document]
ThyWoof: This folder ended up with one patch only so moved back to top folder for now.

SolastaCommunityExpansion/Patches/CustomFeatures/ChangeAbilityCheck/RulesetCharacterPatcher.cs
ThyWoof: We cannot do that in RollDie for ability checks due to passive checks as RollDie only gets called if not passive. RollDie also gets called twice on ResolveContestCheck which makes it a bit harder to store the context.

This particular transpiler is replacing method calls which is much safer than looking for a particular code sequence. It's also pushing the arguments from the original methods to the stack vs local variables. The later can be an issue when we deal with debug and production builds (I put warning comments all over the code for these cases, fyi).

We could implement a similar interface on RollDie to allow changing the results and/or parameters to the call.

ChrisPJohn (approved):

PR #895, opened 2022-03-03
Fix ScoutSentinel

PhilPJL:
Used FormatterServices.GetUninitializedObject to create new objects.  Note this create a completely uninitialized object, no default values, everything 0 or null.  Please check if further setup needs to be done.

Slipped in a .Distinct() on caching definitions.


DubhHerder (approved):
Looks ok assuming the items properties appear in game or JSON as intended.

ThyWoof (approved):

PR #900, opened 2022-03-05
Beta diagnostics

PhilPJL:
Added another export button to export all TA blueprints after CE has loaded over the top of the previously exported TA blueprints. That way we can see which have been affected.  Unfortunately in quite a few definitions, taking a copy reorders the feature unlocks although all the same ones are there.  

Spotted the two known issues and fixed properly (I hope).  These issues are in BuildWitchFamiliar and SwitchHelpPower.

Didn't see anything else of concern, but please try the export and check.  Not intending to check in the results of that export.
Ideally would use reflection to do a full compare of 2 definitions and export changed defs.


SolastaCommunityExpansion/Models/PowersContext.cs
```diff
@@ -50,14 +51,17 @@ internal static void SwitchHelpPower()
 
         private static void LoadHelpPower()
         {
-            var effectDescription = new EffectDescription();
-
-            effectDescription.Copy(TrueStrike.EffectDescription);
+            var effectDescription = TrueStrike.EffectDescription.Copy();
```
ChrisPJohn, line 54: Should we be migrating this to use the effect description builder?

SolastaCommunityExpansion/Models/PowersContext.cs
```diff
@@ -50,14 +51,17 @@ internal static void SwitchHelpPower()
...
             effectDescription.SetTargetType(RuleDefinitions.TargetType.Individuals);
-            effectDescription.EffectForms[0].ConditionForm.ConditionDefinition.GuiPresentation.SetDescription("Condition/&HelpActionDescription");
-            effectDescription.EffectForms[0].ConditionForm.ConditionDefinition.GuiPresentation.SetTitle("Condition/&HelpActionTitle");
+
+            var helpPowerCondition = ConditionDefinitionBuilder
+                .Create(DatabaseHelper.ConditionDefinitions.ConditionTrueStrike, "ConditionHelpPower", DefinitionBuilder.CENamespaceGuid)
+                .SetGuiPresentation("HelpAction", Category.Condition)
+                .AddToDB();
+
+            effectDescription.EffectForms[0].ConditionForm.ConditionDefinition = helpPowerCondition;
```
ChrisPJohn, line 64: Does the EffectDescription.Copy do a deep or shallow copy? Does modifying the EffectForms here change the TrueStrike effect forms?

ChrisPJohn (commented):
Some questions on the help action changes.

SolastaCommunityExpansion/Models/PowersContext.cs
PhilPJL: EffectDescription.Copy does a deep-ish copy. Clones EffectForm and Form, but doesn't clone referenced definitions, just copies the reference.  Doesn't clone GuiPresentation.

SolastaCommunityExpansion/Models/PowersContext.cs
PhilPJL: Yes will do.

ChrisPJohn (approved):

PR #901, opened 2022-03-05
Now only show square/round cylinder.  

PhilPJL:
TA has fixed Grease, Entangle, Black Tentacles so they don't affect flying creatures.
All we need to do is change the geometric cursor to show cylinder/square cylinder.


ThyWoof (changes requested):
Can this be one setting and add spike growth to the list on the other setting?

PhilPJL:
> Can this be one setting and add spike growth to the list on the other setting?

Done.  Moved setting to House Rules.

ThyWoof (approved):

PR #905, opened 2022-03-07
Action Enumerators Context

ThyWoof:
Adding this initially as a tool for modders to get logging on Battle Service methods called from ActionMagicEffect and ActionAttack;

These patch the constructor that creates the internal class returned from these yield calls. In practice they work as Prefix to IEnumerator Method();

Can be used to get context on specific steps or change the call parameters as required;


ThyWoof:
image

image

ThyWoof:
image
multiple targets on spell

PhilPJL (commented):
I'm going to need more time to understand what you're doing :)

The code definitely looks good, and produces some useful logging. Should they be warnings or info?

I haven't tried this but is an alternative less intrusive way to achieve the same thing to do something like this?

image

SolastaCommunityExpansion/Settings.cs
```diff
@@ -27,6 +27,8 @@ public class Settings : UnityModManager.ModSettings
         // Diagnostics
         //
 
+        public bool EnableBattleServiceEventLogs { get; set; } = true;
```
PhilPJL, line 30: Should we default this to false?

PR #912, opened 2022-03-09
1.3.20 custom features patches fixes

ThyWoof:
Use hero from method parameter for better multiplayer compatibility.


ThyWoof:
CharacterBuildingManager now has as CurrentLocalHeroCharacter that might not be set if under a MP scenario. There is a new CharacterHeroBuildingData that keeps this building data. We should whenever possible get the reference from there.

ThyWoof:
@ChrisPJohn , these patches are yours

ThyWoof:
in fact one is mine ;-).

PR #921, opened 2022-03-10
Beta spell fixes

### LINKED ISSUE #742: Feature Request: Make 5th+ Level Bestow Curse Not Require Concentration (Per Tabletop Rules)
In tabletop, Bestow Curse does not require concentration if cast with a 5th-level spell slot or higher. In Solasta this does not seem to be the case. It would be nice if it matched tabletop (or if there an option to make it match by preference), as this is one of the primary selling points of the spell.

PhilPJL:
1) Implements #742.  
2) Remove invalid AlteredDuration from ProtectionFromEnergy/Poison spells.
3) Show 'Concentration' tag on Blade Barrier UI.
4) Fix invalid AlteredDuration on DominateBeast/Person/Monster


ThyWoof (approved):

DubhHerder:
changes to SRD spells look ok.

PR #922, opened 2022-03-10
Allow Shift-R to rotate item

PhilPJL:
Tiny QOL.  Allow Shift-R to rotate say a wall-blocker, so when dropping loads of wall blockers we can rotate the dragged blocker without releasing shift.

### DIFF
--- a/SolastaCommunityExpansion/Models/GameUiContext.cs
+++ b/SolastaCommunityExpansion/Models/GameUiContext.cs
@@ -32,6 +32,9 @@ namespace SolastaCommunityExpansion.Models
         {
             var inputService = ServiceRepository.GetService<IInputService>();
 
+            // Dungeon Maker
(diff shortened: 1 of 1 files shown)


SolastaCommunityExpansion/Models/GameUiContext.cs
```diff
@@ -32,6 +32,9 @@ internal static void Load()
         {
             var inputService = ServiceRepository.GetService<IInputService>();
 
+            // Dungeon Maker
+            inputService.RegisterCommand(InputCommands.Id.EditorRotate, 114, (int)KeyCode.LeftShift, -1, -1, -1, -1);
```
ThyWoof, line 36: Can we use Keycode instead of 114?

ThyWoof (changes requested):

SolastaCommunityExpansion/Models/GameUiContext.cs
PhilPJL: Done

ThyWoof (approved):

PR #923, opened 2022-03-11
Make sure the spell UI works properly through transitions in and out …

ChrisPJohn:
…of battle.

### DIFF
--- a/SolastaCommunityExpansion/Patches/GameUiLocation/SpellSelectionPanelPatcher.cs
+++ b/SolastaCommunityExpansion/Patches/GameUiLocation/SpellSelectionPanelPatcher.cs
@@ -11,7 +11,6 @@ namespace SolastaCommunityExpansion.Patches.GameUiLocation
     internal static class SpellSelectionPanelPatcher
     {
         private static readonly List<RectTransform> spellLineTables = new List<RectTransform>();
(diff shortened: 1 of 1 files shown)


PhilPJL:
Works better but there are missing spell level labels - missing I, III and IV here - this is out of battle

image

and here with different spell selection missing I, IV and VI - this is the same in and out of battle

image

Same issue without mod (different spells again) but VI is missing

image

PhilPJL:
Tried on DLC.  The spell level labels seemed to be ok out of battle, but got exception in battle.
image

ChrisPJohn:
> Now works fine. However there is a problem with missing spell level labels - I, III and IV here - this is out of battle
> 
> image
> 
> and here with different spell selection missing I, IV and VI. Exactly the same in and out of battle.
> 
> image
> 
> Same issue without mod (different spells again) but VI is missing
> 
> image

I'm not sure why the spell level numbers sometimes don't show up. As you noted it happens without the mod (though maybe a little less). I don't think anything in this change effects those numbers.

ChrisPJohn:
> Tried on DLC. The spell level labels seemed to be ok out of battle, but got exception in battle. image

After this is in dev I can investigate what might be going on in beta. I haven't downloaded the latest patch yet, so I don't know what they've changed there.

ThyWoof (approved):

PR #938, opened 2022-03-14
Beta refactor

PhilPJL:
More refactoring.
Mostly LifeTransmuter and SpellMaster.


SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerPoolModifierBuilder.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        public FeatureDefinitionPowerPoolModifierBuilder Configure(
+            int powerPoolModifier, RuleDefinitions.UsesDetermination usesDetermination,
+            string usesAbilityScoreName, FeatureDefinitionPower poolPower)
+        {
+            Definition.SetFixedUsesPerRecharge(powerPoolModifier);
+            Definition.SetUsesDetermination(usesDetermination);
+            Definition.SetUsesAbilityScoreName(usesAbilityScoreName);
+            // This is just an activation time that should not be shown in the UI.
+            Definition.SetActivationTime(RuleDefinitions.ActivationTime.Permanent);
```
ChrisPJohn, line 33: All of these hardcoded values should be set in the constructor. In particular making sure effect description is set and that cost per use is 1 (though really everything that's hardcoded should just be set in the constructor).

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerPoolModifierBuilder.cs
```diff
@@ -0,0 +1,75 @@
...
+            // Math for usage gets weird if this isn't 1.
+            Definition.SetCostPerUse(1);
+            // The game throws an exception if there is no effect description.
+            Definition.SetEffectDescription(new EffectDescription());
+            Definition.SetOverriddenPower(Definition);
+
+            Definition.PoolPower = poolPower;
```
ChrisPJohn, line 40: It should not be possible to construct a power pool modifier without providing a pool. I'd prefer either it is required in the constructor or we have validation to make sure it's been set.

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerSharedPoolBuilder.cs
```diff
@@ -13,6 +14,16 @@ public class FeatureDefinitionPowerPoolBuilder : FeatureDefinitionPowerBuilder<F
...
             RuleDefinitions.UsesDetermination usesDetermination, string usesAbilityScoreName,
             RuleDefinitions.RechargeRate recharge, GuiPresentation guiPresentation) : base(name, guid)
+        {
+            Configure(usesPerRecharge, usesDetermination, usesAbilityScoreName, recharge);
+
+            Definition.SetGuiPresentation(guiPresentation);
+        }
+
+        // TODO: move this to FeatureDefinitionPowerBuilder as ConfigurePowerPool()?
+        public FeatureDefinitionPowerPoolBuilder Configure(int usesPerRecharge,
```
ChrisPJohn, line 24: Similar to other builder changes, I don't like that there is now a way to construct this without some of the values getting set (for example cost per use).

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerSharedPoolBuilder.cs
```diff
@@ -64,6 +78,15 @@ public class FeatureDefinitionPowerSharedPoolBuilder : FeatureDefinitionPowerBui
             RuleDefinitions.ActivationTime activationTime, int costPerUse,
             bool proficiencyBonusToAttack, bool abilityScoreBonusToAttack, string abilityScore,
             EffectDescription effectDescription, GuiPresentation guiPresentation, bool uniqueInstance) : base(name, guid)
+        {
+            Configure(poolPower, recharge, activationTime, costPerUse, proficiencyBonusToAttack, abilityScoreBonusToAttack, abilityScore, effectDescription, uniqueInstance);
+
+            Definition.SetGuiPresentation(guiPresentation);
+        }
+
+        public FeatureDefinitionPowerSharedPoolBuilder Configure(FeatureDefinitionPower poolPower,
```
ChrisPJohn, line 87: I don't like that there is now a path to construct that allows some critical fields (uses determination, shared pool) to not be set.

SolastaCommunityExpansion/Subclasses/Wizard/SpellMaster.cs
```diff
@@ -101,61 +101,32 @@ internal SpellMaster()
...
         {
             if (Main.Settings.EnableUnlimitedArcaneRecoveryOnWizardSpellMaster)
             {
-                BonusRecovery.SetGuiPresentation(
-                    GuiPresentationBuilder.Build(
-                        "MagicAffinitySpellMasterRecoveryUnlimited",
-                        Category.Subclass, PowerWizardArcaneRecovery.GuiPresentation.SpriteReference));
-                BonusRecovery.SetCostPerUse(0);
-                BonusRecovery.SetRechargeRate(RuleDefinitions.RechargeRate.AtWill);
+                BonusRecovery
```
ChrisPJohn, line 119: This might need a check that the BonusRecovery has been initialized.

ChrisPJohn (changes requested):

SolastaCommunityExpansion/Subclasses/Wizard/SpellMaster.cs
PhilPJL: I think all definitions have been created before anyone can view the settings page.

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerPoolModifierBuilder.cs
PhilPJL: Added validation

SolastaCommunityExpansion/Subclasses/Wizard/SpellMaster.cs
ThyWoof: Settings won’t render until Main.Enabled equals true which happens at the very end of our init code.

SolastaCommunityExpansion/Builders/ConditionDefinitionBuilder.cs
```diff
@@ -207,10 +206,28 @@ public TBuilder SetDuration(RuleDefinitions.DurationType type, int duration = 0,
...
-        // TODO: add more methods as required (and that aren't delegating property setters to add value)
+        public TBuilder Configure(RuleDefinitions.DurationType durationType, int durationParameter,
+            bool silent, IEnumerable<FeatureDefinition> conditionFeatures)
+        {
+            Definition.Features.AddRange(conditionFeatures);
+            Definition.SetConditionType(RuleDefinitions.ConditionType.Beneficial);
+            Definition.SetAllowMultipleInstances(false);
+            Definition.SetDurationType(durationType);
+            Definition.SetDurationParameter(durationParameter);
+            Definition.SetConditionStartParticleReference(new AssetReference());
```
ChrisPJohn, line 220: Should these asset references be set in the initialise?

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerBuilder.cs
```diff
@@ -70,13 +70,30 @@ protected FeatureDefinitionPowerBuilder(TDefinition original, string name, strin
...
+
+            if (Definition.EffectDescription == null)
+            {
+                // The game throws an exception if there is no effect description.
+                Definition.SetEffectDescription(new EffectDescription());
+            }
+
+            if (IsNew)
+            {
+                // This is just an activation time that should not be shown in the UI.
```
ChrisPJohn, line 85: The activation time and costPerUse defaults here were only supposed to be for the power pool related builders that were previously setting them. I don't think they make sense in the base power builder.

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerPoolModifierBuilder.cs
```diff
@@ -0,0 +1,70 @@
...
+     * at 4 different character levels you need to create 4 different FeatureDefinitionPowerPoolModifier.
+     */
+    public class FeatureDefinitionPowerPoolModifierBuilder : FeatureDefinitionPowerBuilder<FeatureDefinitionPowerPoolModifier, FeatureDefinitionPowerPoolModifierBuilder>
+    {
+        internal override void Validate()
+        {
+            base.Validate();
+
+            Preconditions.IsNotNull(Definition.PoolPower, $"{GetType().Name}[{Definition.Name}].PoolPower is null.");
+            // TODO: more validation?
```
ChrisPJohn, line 21: We should probably validate that costPerUse is 1 for the modifier.

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerSharedPoolBuilder.cs
```diff
@@ -1,90 +1,89 @@
...
-            Definition.SetRechargeRate(recharge);
-            // The game throws an exception if there is no effect description.
-            Definition.SetEffectDescription(new EffectDescription());
-            Definition.SetOverriddenPower(Definition);
-            Definition.SetGuiPresentation(guiPresentation);
+            base.Initialise();
+
+            // We set uses determination to fixed because the code handling updates needs that.
+            Definition.SetUsesDetermination(RuleDefinitions.UsesDetermination.Fixed);
+            // Recharge rate probably shouldn't be in here, but for now leave it be because there is already usage outside of this mod.
```
ChrisPJohn, line 17: Why is this comment here now?

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerSharedPoolBuilder.cs
```diff
@@ -1,90 +1,89 @@
...
-            Definition.SetUsesAbilityScoreName(usesAbilityScoreName);
-            // This is just an activation time that should not be shown in the UI.
-            Definition.SetActivationTime(RuleDefinitions.ActivationTime.Permanent);
-            // Math for usage gets weird if this isn't 1.
-            Definition.SetCostPerUse(1);
-            // The game throws an exception if there is no effect description.
-            Definition.SetEffectDescription(new EffectDescription());
-            Definition.SetOverriddenPower(Definition);
-            Definition.SetGuiPresentation(guiPresentation);
+            base.Validate();
```
ChrisPJohn, line 22: This should validate that cost per use is 1 and uses determination is fixed.

ChrisPJohn (changes requested):

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerSharedPoolBuilder.cs
PhilPJL: done

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerSharedPoolBuilder.cs
PhilPJL: cut and paste error

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerPoolModifierBuilder.cs
PhilPJL: done

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerBuilder.cs
PhilPJL: moved back

SolastaCommunityExpansion/Builders/ConditionDefinitionBuilder.cs
PhilPJL: Yes, they were already being done in ctors.  Added same to Initialise.

SolastaCommunityExpansion/Builders/Features/FeatureDefinitionPowerSharedPoolBuilder.cs
ChrisPJohn: So checking the old code outside of the git diff (because that's a nightmare). I was wrong about cost per use needing to be 1. However we really should validate that usesDetermination is set to fixed. I know it's in the initialize, but since the full power builder is available here it's quite possible a user could change that.

ChrisPJohn (changes requested):

PhilPJL:
Closed to avoid further conflict.

PhilPJL:
I'll address issue raised previously

SolastaCommunityExpansion/Builders/ConditionDefinitionBuilder.cs
```diff
@@ -25,15 +24,38 @@ public abstract class ConditionDefinitionBuilder<TDefinition, TBuilder> : Defini
         where TDefinition : ConditionDefinition
         where TBuilder : ConditionDefinitionBuilder<TDefinition, TBuilder>
     {
+        protected override void Initialise()
+        {
+            base.Initialise();
+
+            // NOTE: if a constructor is used Initialise won't get called from the DefinitionBuilder base
```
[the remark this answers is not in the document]
PhilPJL: That was my original plan.  I've added a flag to disable the use of standard Create methods so the builder can force the use of more specialized Create method or constructor.

SolastaCommunityExpansion/Builders/ConditionDefinitionBuilder.cs
```diff
@@ -25,15 +24,38 @@ public abstract class ConditionDefinitionBuilder<TDefinition, TBuilder> : Defini
...
         {
             var assetReference = new AssetReference();
 
-            Definition
-                .SetConditionStartParticleReference(assetReference)
-                .SetConditionParticleReference(assetReference)
-                .SetConditionEndParticleReference(assetReference)
-                .SetCharacterShaderReference(assetReference);
+            // TODO: add initializer in extension generation
+            if (Definition.GetField<AssetReference>("conditionStartParticleReference") == null)
```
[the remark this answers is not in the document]
PhilPJL: Added ClearParticleReferences and SetEmptyParticleReferencesWhereNull as extension methods.
Used SetEmptyParticleReferencesWhereNull in builder since we may copy an existing definition.

PhilPJL:
@ChrisPJohn 

added

```
        internal override void Validate()
        {
            base.Validate();

            Preconditions.IsNotNull(Definition.SharedPool, $"FeatureDefinitionPowerSharedPoolBuilder[{Definition.Name}].SharedPool is null.");
            **Preconditions.AreEqual(Definition.UsesDetermination, RuleDefinitions.UsesDetermination.Fixed, $"FeatureDefinitionPowerSharedPoolBuilder[{Definition.Name}].UsesDetermination must be set to Fixed.");**
        }
```

Does your comment above about CostPerUse mean I should remove 

`            Preconditions.AreEqual(Definition.CostPerUse, 1, $"{GetType().Name}[{Definition.Name}].CostPerUse must be set to 1.");
`

from `FeatureDefinitionPowerPoolModifierBuilder `and/or `FeatureDefinitionPowerPoolBuilder`?

ChrisPJohn:
> 

FeatureDefinitionPowerPoolModifierBuilder and FeatureDefinitionPowerPoolBuilder should both still verify cost per use is 1.

FeatureDefinitionPowerSharedPoolBuilder does not need to validate costPerUse (can be any).

ChrisPJohn (approved):

PR #947, opened 2022-03-18
adding extra high CR monsters (from DMPRO)

DubhHerder:


ChrisPJohn (commented):
Generally looks good. I skimmed the larger files, since my understanding is this is content merged over from an existing source.

@PhilPJL should probably take a look since he's been the builder expert recently.

DubhHerder:
The original version of the monster definition builder came was a piece of code when it was written last August. The code might be messy but that's because each attribute required iterative testing to guarantee intended monster behaviour

Both unique monster powers and AI decisions require string parameters to function correct and so no names or strings should changed

It's legacy code so it needed the legacy base definitions builder to be enabled again. 

I didn't have the JSON definitions pointed at the correct folder so I'll update that tomorrow night or Sunday

PhilPJL (commented):
Ok I can't be doing with all of those silly comments.  Instead of re-enabling legacy code why don't you use the new builders?

DubhHerder:
> Ok I can't be doing with all of those silly comments. Instead of re-enabling legacy code why don't you use the new builders?

It's legacy code that was written before the new builders were created. That's why the new builders were not used.

The comments are meant as a reminder to all that making changes without testing is not refactoring.

PhilPJL (approved):
I'll approve, but will create a PR to make BaseDefinitionBuilder obsolete again and use new builders (and check the blueprints don't change).

DubhHerder:
That is not necessary. You can just leave the code as it is. it is functional and has been for months. . 

General Reply: This mod is a collection of work from the Solasta modding community not a single individual and the general philosophy is everything is optional not that everything must conform to one opinion. This legacy code isnt perfect but it does not negatively effect anybody else or their work. There's things I disagree with in the mod but I only comment on things that either interest me or that impact the code I've contributed and am expected to maintain. 

Specific Reply:  monsters are more complex and interdependent than almost any other definition so any changes are much more likely to break them. Not only do they have features,  powers, spelllists, spell casting, etc  all with associated conditions and attribute modifiers similar to PC classes and subclasses they also have unique spells, attacks , legendary actions, savings throws, attribute stats, group attacks, sizes, types, visual models, shaders, particles, sprites, AI decisions, AI packages. All of which I have created or customised for these monsters. If you refractor this, You will break the the monsters. Do not do that. 

Finally, If you change the code then I have no interest in maintaining it. I have limited time, and I cant waste it learning how you will restructure it while trying to fix any potential future bugs.  Please respect that not everyone has the same amount of free time as you might.

PR #948, opened 2022-03-18
remove target filtering tag

DubhHerder:


ChrisPJohn (approved):
I'm approving the change since I trust you tested it. I was expecting a json diff from the definition output (I guess this might be the default though, didn't double check that).

DubhHerder:
I manually checked the generated JSON for the affected powers to confirm the change. The definitions are defaulting to the game folder because I haven't used thus feature before now. I can correct it tomorrow night or Sunday to direct the definitions to the GitHub folder instead so that the change in definitions is captured

PR #955, opened 2022-03-19
Update GuiPresentationBuilder to copy the current GuiPresentation

PhilPJL:
Update GuiPresentationBuilder to copy the definition's GuiPresentation if it has one.

Exported CE blueprints. Several new blueprints. Many blueprints updated with m_SubObjectName = "", m_SubObjectType = "".  Some blueprints now have the sprite reference from the copied definition whereas before it was empty.

The potential issue is that when copying an existing definition with
    `XXXBuilder.Create(original, name, guid).SetGuiPresentation(...)`
this would always set a new GuiPresentation and wouldn't copy the GuiPresentation from 'original'.  Now it does along with value of hidden and sortOrder.

Difference is that previously you would need to explicitly copy these fields, now you need to explicitly remove them.  
I think this is possibly the better approach.  I'll let you decide.


ChrisPJohn:
> Update GuiPresentationBuilder to copy the definition's GuiPresentation if it has one.
> 
> Exported CE blueprints. Several new blueprints. Many blueprints updated with m_SubObjectName = "", m_SubObjectType = "". Some blueprints now have the sprite reference from the copied definition whereas before it was empty.
> 
> The potential issue is that when copying an existing definition with `XXXBuilder.Create(original, name, guid).SetGuiPresentation(...)` this would always set a new GuiPresentation and wouldn't copy the GuiPresentation from 'original'. Now it does along with value of hidden and sortOrder.
> 
> Difference is that previously you would need to explicitly copy these fields, now you need to explicitly remove them. I think this is possibly the better approach. I'll let you decide.

In my opinion a function named SetX should set X to the provided value. Keeping some values when you do a set feels like a hidden trap for modders to fall into.

PhilPJL:
> In my opinion a function named SetX should set X to the provided value. Keeping some values when you do a set feels like a hidden trap for modders to fall into.

Yes I agree.  I did think of adding UpdateGuiPresentation but was trying to make it automatically fix code that has potentially been broken by me changing Title='...', etc to SetGuiPresentation.
I'll change SetGuiPresentation back to just setting a new one and add SetOrUpdate which will update an existing one or create a new one if not present.

PhilPJL:
Added SetOrUpdateGuiPresentation and used that in specific cases.

AbilityCheckAffinityFeatPickPocket and ProficiencyFeatPickPocket still have hidden=false whereas the original copied GuiPresentation has hidden=true.  What value is required?

ChrisPJohn:
> 

I'm not the original author or even the one that ported it over, so I'm just guessing. Based on the original code:
https://github.com/ChrisPJohn/SolastaCommunityExpansion/commit/77b57c9cb0929ae9f22922b84a3e3904243f92e4

I think that would have resulted in hidden=true if the base was hidden=true. At the same time they set title/description, so I'd be surprised if it wasn't displayed at some point.

ChrisPJohn (commented):
Looks good to me, but @DubhHerder and @RealBazou might want to take a look and verify this is changing things back to the original intent.

DubhHerder:
changes look correct

ChrisPJohn (approved):

PR #964, opened 2022-03-20
Zappa Feats

ThyWoof:
changed Feats:

* Metamagic feats now requires Character Level 4 (Heightened requires level 8)
* Nerfed Shady feat with a slower sneak damage progression

added new Feats:

* Brutal Thug - sneak attacks with martial weapons and fighter weapon proficiency
* Metamagic: Heightened - learn Heightened and gain 4 sorcery points
* Arcane / Charismatic / Wise Precision - Power / 2 uses / long rest / give 1 min Int / Cha / Wis bonus to attack and damage.
* Arcane Defense - Like other defenses in the mod

created FeatDefinitionBuilder abstract class and FeatDefinitionBuilder / FeatDefinitionWithPrereqsBuilder implementations.

extended AttributeModifier calculation:

* add new modifier AdditiveAtEnd - allows addition after multiply / multiply by level / multiply by class level (default additive modifier happens before multiplications) [required to allow adding bonus sorcery points to a sorcerer]

moved 2 patches from GameUI/LevelUp into CustomFeatures.

tweaked custom fighting style texts again. Had to split Feat and Fighting styles into 2 terms to achieve desired goal.

sorted translations-en file (will redo translations to other languages when closer to April 14th)


ThyWoof:
image
different titles for custom fighting styles and them as feats.

ThyWoof:
image
these descriptions in one line are a game bug. I shared with TA and got confirmation on a fix.

ThyWoof:
image
image

ThyWoof:
image
Sorcerer Level 4 with 1 metamagic feat. Sorcery Points = 4 (level) + 2 (feat).

image
Sorcerer Level 8 with 2 metamagic feats. Sorcery Points = 8 (level) + 2 (feat) + 4 (Heightened Feat)

SolastaCommunityExpansion/Patches/CustomFeatures/AtttributeModifierBonus/RulesetAttributePatcher.cs
```diff
@@ -0,0 +1,23 @@
...
+using HarmonyLib;
+using SolastaModApi.Extensions;
+
+namespace SolastaCommunityExpansion.Patches.CustomFeatures.AtttributeModifierBonus
+{
+    [HarmonyPatch(typeof(RulesetAttribute), "Refresh")]
+    [SuppressMessage("Minor Code Smell", "S101:Types should be named in PascalCase", Justification = "Patch")]
+    internal static class RulesetAttribute_Refresh
+    {
+        internal static void Postfix(RulesetAttribute __instance, ref int ___currentValue)
```
ChrisPJohn, line 12: I think we can have a simpler patch. So before attributemodifiers are applied, they are generally sorted. With the sort order being determined by a comparison between the operation enum values (and then several tie break checks). 

Since the value you chose for the enum means it will get sorted to the end of the list, you could add a postfix to RulesetAttributeModifier.ApplyOnValue and just do the addition there.

ChrisPJohn (changes requested):

SolastaCommunityExpansion/Patches/CustomFeatures/AtttributeModifierBonus/RulesetAttributePatcher.cs
ThyWoof: ok.

ThyWoof:
@ChrisPJohn : applied the requested changes and validated pool calculation was still correct. same on pre reqs. still working after refactor.

ThyWoof:
The Feat builder can now accept one or more validation delegates for more complex scenarios.

SolastaCommunityExpansion/Feats/FeatsValidations.cs
```diff
@@ -0,0 +1,121 @@
...
+
+namespace SolastaCommunityExpansion.Feats
+{
+    internal static class FeatsValidations
+    {
+        //
+        // validation routines for FeatDefinitionWithPrereqsBuilder
+        //
+
+        internal static bool ValidateMinCharacterLevel4(
```
ChrisPJohn, line 11: The reason I suggested an abstract class where you provide the function as part of the class was you could add relevant values to the class as fields. So the you'd be able to have 1 new class for level prereqs, and set the level to check on that class.

SolastaCommunityExpansion/Feats/FeatsValidations.cs
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            var levelText = isLevelValid ? "8" : Gui.Colorize("8", "EA7171");
+
+            prerequisiteOutput += Gui.Format("Tooltip/&FeatPrerequisiteLevelFormat", levelText);
+
+            return isLevelValid;
+        }
+
+        internal static bool ValidateNotBarbarian(
```
ChrisPJohn, line 49: For all of these feats that require not being a character class what happens with multiclassing?

SolastaCommunityExpansion/Patches/CustomFeatures/FeatsPrerequisites/GuiFeatDefinitionPatcher.cs
```diff
@@ -19,16 +19,20 @@ internal static class GuiFeatDefinition_IsFeatMacthingPrerequisites
...
         {
-            foreach (var featureDefinition in feat.Features.OfType<IValidateFeatPrerequisites>())
+            if (feat is not FeatDefinitionWithPrereqs featDefinitionWithPrereqs || featDefinitionWithPrereqs.IsFeatMacthingPrerequisites == null)
             {
-                
-                var result = featureDefinition.IsFeatMacthingPrerequisites(feat, hero, ref prerequisiteOutput);
+                return;
+            }
+            
+            foreach (IsFeatMacthingPrerequisites isFeatMacthingPrerequisites in featDefinitionWithPrereqs.IsFeatMacthingPrerequisites.GetInvocationList())
```
ChrisPJohn, line 27: I don't believe this needs to be a list. It's a function that can be defined by the modder. If they want to check multiple things in the one function they can.

ChrisPJohn (changes requested):

SolastaCommunityExpansion/Feats/FeatsValidations.cs
ThyWoof: won't cover MC use case here. if someone takes a class after the feat that's fine.

SolastaCommunityExpansion/Feats/FeatsValidations.cs
ThyWoof: I followed the pattern I found on FeatureDefinitionConditionAffinityBuilder. It's easier when I have some code I can read as my C# knowledge isn't that sharp.

SolastaCommunityExpansion/Patches/CustomFeatures/FeatsPrerequisites/GuiFeatDefinitionPatcher.cs
ThyWoof: There will a lot of repetition code in validations otherwise. I tested this scenario and it works fine. Tried to be modular here.

SolastaCommunityExpansion/Feats/FeatsValidations.cs
ThyWoof: If you could point me out to some full examples...

ThyWoof:
image
@ChrisPJohn : here is an example where having modular validation routines make sense. I want Brutal Thug to only be allowed after level 4 but only if hero has sneak attacks. Added 2 validations to this feat.

ThyWoof:
I also renamed to FeatDefinitionCustom so if we ever need more custom behavior we keep extending this class and corresponding builder.

SolastaCommunityExpansion/Feats/FeatsValidations.cs
```diff
@@ -10,42 +11,54 @@ internal static class FeatsValidations
         // validation routines for FeatDefinitionCustomBuilder
         //
 
-        internal static bool ValidateMinCharacterLevel4(
-            FeatDefinition _,
-            RulesetCharacterHero hero,
-            ref string prerequisiteOutput)
+        internal static IsFeatMacthingPrerequisites ValidateMinCharLevel(int minCharLevel)
```
PhilPJL, line 14: Spelling 'IsFeatMatching'

PhilPJL:
Validation may fit in with what was proposed on #938 which I closed.  May be worth re-opening.

SolastaCommunityExpansion/Builders/FeatDefinitionCustomBuilder.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+
+namespace SolastaCommunityExpansion.Builders
+{
+    public delegate bool IsFeatMatchingPrerequisites(
+          FeatDefinition feat,
+          RulesetCharacterHero hero,
+          ref string prerequisiteOutput);
+
+    public class FeatDefinitionCustom : FeatDefinition
```
PhilPJL, line 10: Is this the right name?  'Custom' is a bit generic, what is this feat definition's purpose?  Should it be FeatDefinitionWithValidation (something like that?)

SolastaCommunityExpansion/Feats/FeatsValidations.cs
```diff
@@ -0,0 +1,93 @@
...
+
+namespace SolastaCommunityExpansion.Feats
+{
+    internal static class FeatsValidations
+    {
+        //
+        // validation routines for FeatDefinitionCustomBuilder
+        //
+
+        internal static IsFeatMatchingPrerequisites ValidateMinCharLevel(int minCharLevel)
```
PhilPJL, line 14: I like it but I wouldn't have used (or thought of tbh) delegates, I would have used Func<def, hero, (bool, string)>.  
The string concatenation doesn't belong in the validators they should return the result (bool, string) only.  There should then be a helper method somewhere to concat the strings.  `string.Join("\n", results)`

SolastaCommunityExpansion/Builders/FeatDefinitionCustomBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        public IsFeatMatchingPrerequisites IsFeatMatchingPrerequisites;
```
PhilPJL, line 12: make it a property.  Not good practice to have public fields.

PhilPJL: or is it for delegates?

SolastaCommunityExpansion/Patches/CustomFeatures/FeatsPrerequisites/GuiFeatDefinitionPatcher.cs
```diff
@@ -4,14 +4,37 @@
...
-    internal static class GuiFeatDefinition_IsFeatMacthingPrerequisites
+    internal static class GuiFeatDefinition_IsFeatMatchingPrerequisites
     {
+        internal static void Postfix(
+            ref bool __result,
+            FeatDefinition feat,
+            RulesetCharacterHero hero,
+            ref string prerequisiteOutput)
+        {
+            if (feat is not FeatDefinitionCustom FeatDefinitionCustom || FeatDefinitionCustom.IsFeatMatchingPrerequisites == null)
```
PhilPJL, line 22: lower case please `featDefinitionCustom `

SolastaCommunityExpansion/Patches/CustomFeatures/FeatsPrerequisites/GuiFeatDefinitionPatcher.cs
```diff
@@ -4,14 +4,37 @@
...
+            {
+                return;
+            }
+            
+            foreach (IsFeatMatchingPrerequisites IsFeatMatchingPrerequisites in FeatDefinitionCustom.IsFeatMatchingPrerequisites.GetInvocationList())
```
PhilPJL, line 27: I guess I would put this logic in FeatDefinitionCustom 

```
void Test(ref bool __result, ref string prerequisiteOutput)
{
	var feat = new FeatDefinitionCustom();

// and call it here
	var (result, output) = feat.Validate();
	
	__result = result;
	prerequisiteOutput = output;
}

class FeatDefinitionCustom
{
	public (bool result, string output) Validate()
	{
		// call all validators 
		// concat output appropriately
		// return result appropriately
		
		return (true, "It's all good");
	}
}
```

ThyWoof:
> I also renamed to FeatDefinitionCustom so if we ever need more custom behavior we keep extending this class and corresponding builder.

It was named FeatDefinitionWithPrereqs before. I can revert that.

SolastaCommunityExpansion/Builders/FeatDefinitionCustomBuilder.cs
ThyWoof: it's a delegate. gotta be public.

SolastaCommunityExpansion/Feats/FeatsValidations.cs
ThyWoof: using delegates to have modular validations. I'm also following the same pattern on game code that call these checks in sequence, updating the ref everytime.

SolastaCommunityExpansion/Patches/CustomFeatures/FeatsPrerequisites/GuiFeatDefinitionPatcher.cs
ThyWoof: ok

SolastaCommunityExpansion/Patches/CustomFeatures/FeatsPrerequisites/GuiFeatDefinitionPatcher.cs
ThyWoof: check GuiFeatDefinition.IsFeatMacthingPrerequisites to understand why I chose this pattern.

SolastaCommunityExpansion/Builders/FeatDefinitionCustomBuilder.cs
ThyWoof: It was named FeatDefinitionWithPrereqs before. I can revert that.

SolastaCommunityExpansion/Feats/FeatsValidations.cs
ThyWoof: I disagree with string not belonging on validations. The way I wrote make validation routines self contained. You can mix and match more than one on same Feat.

ThyWoof:
@PhilPJL / @ChrisPJohn : applied suggestions by Imp. Code is cleaner now on validation routines and it seems to be faster as it joins all results together.

ThyWoof:
image
@ChrisPJohn : I ended up setting WarCaster to require spell casting. There was a bug on the code removing this validation that got fixed.

SolastaCommunityExpansion/Patches/CustomFeatures/FeatDefinitionCustomCode/GuiFeatDefinitionPatcher.cs
```diff
@@ -56,9 +52,7 @@ internal static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruct
         private static int CanCastSpells(RulesetCharacterHero hero)
         {
             return Main.Settings.EnableFirstLevelCasterFeats
-                // Replace call to RulesetCharacterHero.SpellRepertores.Count with Count list of FeatureCastSpell 
-                // which are registered before feat selection at lvl 1
-                ? hero.EnumerateFeaturesToBrowse<FeatureDefinitionCastSpell>().Count
+                ? 1
```
ChrisPJohn, line 55: If we just wanted to have this always on, the patch could be a lot simpler. Why are you making this change?

ChrisPJohn (changes requested):

PR #971, opened 2022-03-23
More Feats

ThyWoof:


ThyWoof:
This request touches translations which is holding other activities I was willing to do. Appreciated if we get into a consensus here.

SolastaCommunityExpansion/Feats/FeatsValidations.cs
```diff
@@ -0,0 +1,68 @@
...
+
+namespace SolastaCommunityExpansion.Feats
+{
+    internal static class FeatsValidations
+    {
+        //
+        // validation routines for FeatDefinitionWithPrerequisites
+        //
+
+        internal static Func<FeatDefinition, RulesetCharacterHero, (bool result, string output)> ValidateMinCharLevel(int minCharLevel)
```
PhilPJL, line 14: minCharLevel isn't used.

SolastaCommunityExpansion/Patches/CustomFeatures/FeatDefinitionWithPrereqs/GuiFeatDefinitionPatcher.cs
```diff
@@ -0,0 +1,28 @@
...
+            ref string prerequisiteOutput)
+        {
+            if (feat is not FeatDefinitionWithPrerequisites featDefinitionWithPrerequisites || featDefinitionWithPrerequisites.Validators.Count == 0)
+            {
+                return;
+            }
+
+            var (result, output) = featDefinitionWithPrerequisites.Validate(featDefinitionWithPrerequisites, hero);
+
+            __result &= result;
```
PhilPJL, line 24: I think that should be ___result &&= result;_

SolastaCommunityExpansion/Feats/FeatsValidations.cs
ThyWoof: it is on 22 and 26 but indeed there was still a hard coded 4 there on 18.

PhilPJL: that's what I meant

ThyWoof: I read the comment on &= but you removed. Confirmed &= was working but changed to A = A && B.

PhilPJL (approved):
I'm ok with this approach.
I haven't tested the feats in game - no time.

PR #986, opened 2022-03-25
Update blueprints

PhilPJL:
Updated blueprints after recent PRs.  Mostly adds with a couple of modifications.
Not sure about these
image

which result from this change in FightingStyleFeats.cs
image


ThyWoof:
These changes were made to have consistency when offering FS as feats. This was discussed with @ChrisPJohn on another request.

ThyWoof (approved):

PR #988, opened 2022-03-27
Non stacking AC with Mage Armor (Defense Feats)

ThyWoof:
* added situational context to all 3 defense feats to avoid stacking with mage armor
* added a new builder method on FeatureDefinitionAttributeModifier


PhilPJL (approved):
Looks ok to me.  Happy to let Chris double check.

PR #1009, opened 2022-03-31
add Gnome

ThyWoof:


SolastaCommunityExpansion/Models/RacesContext.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Reflection;
+using static SolastaModApi.DatabaseHelper;
+
+namespace SolastaCommunityExpansion.Models
+{
+    public class RacesContext
+    {
+        internal static CharacterRaceDefinition GnomeRace { get; private set; } = BuildGnome();
+
+        private static T MemberwiseClone<T>(T base_object)
```
PhilPJL, line 13: I don't think this is doing what you expect.  It's a shallow copy so you may find modifying collections in the copy modify the original.

Try this in LinqPad

```
void Main()
{
	var t = new Test { Number = 1, Items = new List<string> { "one", "two" }};

	t.Number.Dump("orig");
	string.Join(",", t.Items).Dump("orig");

// take copy
	var t2 = MemberwiseClone(t);

// modify original
	t.Number = 5;
	t.Items.Add("three");

// we see change to Items
	t2.Number.Dump("copy");
	string.Join(",", t2.Items).Dump("copy");
}

class Test
{
	public int Number {get;set;}
	public List<string> Items {get;set;}
}
// Define other methods and classes here
private static T MemberwiseClone<T>(T base_object)
{
	MethodInfo memberwiseCloneMethod = typeof(T).GetMethod("MemberwiseClone", BindingFlags.Instance | BindingFlags.NonPublic);
	return (T)memberwiseCloneMethod.Invoke(base_object, null);
}
```

SolastaCommunityExpansion/Models/RacesContext.cs
```diff
@@ -0,0 +1,115 @@
+﻿using SolastaCommunityExpansion.Builders;
+using SolastaCommunityExpansion.Builders.Features;
+using SolastaModApi.Extensions;
+using System.Reflection;
+using static SolastaModApi.DatabaseHelper;
+
+namespace SolastaCommunityExpansion.Models
+{
+    public class RacesContext
```
PhilPJL, line 9: public static

SolastaCommunityExpansion/Models/RacesContext.cs
PhilPJL: See https://docs.microsoft.com/en-us/dotnet/api/system.object.memberwiseclone?view=net-6.0

PhilPJL: Maybe Object.Instantiate would work as a deep copy.

PhilPJL (changes requested):
Clone method is faulty

SolastaCommunityExpansion/Models/RacesContext.cs
ThyWoof: this came from migration. Object.Instantiate won't work on this class. Any suggestions?

PhilPJL: Manually create a copy method.

var orig = CharacterRaceDefinitions.Halfling.RacePresentation;
var rp = new RacePresentation();

rp.SetMaleBeardShapeOptions(orig.MaleBeardShapeOptions);
... etc

ThyWoof: implemented https://github.com/Burtsev-Alexey/net-object-deep-copy

PhilPJL (approved):
Only issues I can think of are
1) the gnome and warlock images are high res and take up about 2Mb. The final .dll is <5Mb  total including the images.  Maybe they could be shrunk down.
2) was concerned the new .Copy could conflict with other .Copy methods but it doesn't seem to.  Maybe rename it DeepCopy?

PhilPJL (changes requested):
I'm thinking Copy methods should be called DeepCopy to avoid accidentally using this Copy when there's a Copy extension method on a blueprint.

Also DeepCopy should throw if attempting to copy a UnityEngine.Object.  A DeepCopy of a blueprint/Unity object may cause problems and we should use Object.Instantiate etc in those cases.

ThyWoof:
renamed to DeepCopy and added check on unity objects

PhilPJL (approved):

PR #1018, opened 2022-04-01
fix Warlock spell list name

ThyWoof:


SolastaCommunityExpansion/Spells/SRDSpells.cs
```diff
@@ -71,29 +71,29 @@ internal static void Register()
         {
             // 7th level
             SpellsContext.RegisterSpell(DivineWord, isFromOtherMod: false, SpellsContext.NOT_IN_MIN_SET, "SpellListCleric");
-            SpellsContext.RegisterSpell(FingerOfDeath, isFromOtherMod: false, "SpellListClassWarlock", "SpellListWizard", "SpellListSorcerer");
```
PhilPJL, line 74: Maybe these spell list names should be consts somewhere?
const string WarlockSpellListName = "ClassWarlockSpellList"

PhilPJL (approved):

PR #1023, opened 2022-04-01
fix bug with Damage Force particle parameters

ThyWoof:
getting a null exception on this method when DamageForce is passed

### DIFF
--- a/SolastaCommunityExpansion/Patches/Bugfix/ConditionDefinitionPatcher.cs
+++ b/SolastaCommunityExpansion/Patches/Bugfix/ConditionDefinitionPatcher.cs
@@ -1,0 +1,23 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using HarmonyLib;
+
+namespace SolastaCommunityExpansion.Patches.Bugfix
+{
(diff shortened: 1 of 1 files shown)


SolastaCommunityExpansion/Patches/Bugfix/ConditionDefinitionPatcher.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using HarmonyLib;
+
+namespace SolastaCommunityExpansion.Patches.Bugfix
+{
+    [HarmonyPatch(typeof(ConditionDefinition), "GetParticleParametersFromAncestryDamageType")]
```
PhilPJL, line 6: when/where is the exception thrown?  Is this a TA bug or something we did?

PhilPJL: Ok, obvious if I look at their code.  They're actually doing some checking, who would expect that :)

ThyWoof:
this is only happening with MC heroes. I'll add this patch under MC for now.

PR #1028, opened 2022-04-01
Beta tidy

PhilPJL:
First pass tidy up to eliminate warnings.
Fix SetSpellsAtLevel so it actually works.
Only one BP changed ClassWarlockSpellList which was result of another PR (I think).


ThyWoof:
Confirmed the blueprint change came from an earlier PR you merged yesterday.

ThyWoof (approved):

PR #1029, opened 2022-04-01
reduce resource images filesize / tweak credits / update json URLs

ThyWoof:


PhilPJL (changes requested):
Change the CreateAssetReferenceSprite calls with the new image sizes

ThyWoof:
Image sizes haven't changed. Only filesize.

PhilPJL (approved):
Ok. I was thinking they could be lower resolution. If they need to be the original resolution I didn't have any objection to the file size

PR #1033, opened 2022-04-02
rollback changes using new spell list builder

ThyWoof:
@PhilPJL : for some reason the SetAtSpellLevel is screwing up the spell offering (i.e.: getting sleep as cantrip). Rolled back the changes and ensured all 4 subs are working. I reviewed all other classes and only Warlock and Witch main spell list use SetAtSpellLevel. They work fine as they have a non empty cantrip list.


PhilPJL (commented):
I would prefer to fix it properly

PR #1036, opened 2022-04-02
Beta tidy & fix spell list builder

PhilPJL:


ThyWoof:
image
this is incorrect. when used on subs the spell list builder is offering level 1 spells as cantrips [sleep]

ThyWoof (approved):

PR #1041, opened 2022-04-02
add Bolgrif race

ThyWoof:


SolastaCommunityExpansion/Models/RacesContext.cs
```diff
@@ -8,16 +8,120 @@ namespace SolastaCommunityExpansion.Models
...
             //
+            _ = BolgrifRace;
             _ = GnomeRace;
         }
 
+        internal static CharacterRaceDefinition BuildBolgrif()
+        {
+            var bolgrifSpriteReference = Utils.CustomIcons.CreateAssetReferenceSprite("Bolgrif", Properties.Resources.Bolgrif, 1024, 512);
+
+            var bolgrifAbilityScoreModifier1 = FeatureDefinitionAttributeModifierBuilder
```
PhilPJL, line 28: There's no requirement to use specific guids. You can use DefinitionBuilder.CeNamespaceGuid for everything if you want.

ThyWoof: Used them to match extra content guids

PhilPJL (changes requested):
Can you add the blueprints as well?

PhilPJL (approved):

PR #1073, opened 2022-04-08
Guid refactor

ThyWoof:
@PhilPJL : I refactored the GUID string definition on settings. Getting an error with SpellDefinitionBuilder. Sharing this before you start refactoring same. Can you take a look on this error?


PhilPJL (changes requested):
This is the first time those specific Create/Constructors have been used,
like  `XXXDefinitionBuilder.Create("<name>"); XXXDefinitionBuilder.Create(original, "<name>");`

Using Create like that prepends -CE- to the name _and_ auto creates a `GuiPresentation`.
If you're setting a `GuiPresentation `anyway then you can optionally do `XXXDefinitionBuilder.Create("<name>", false);` to avoid creating the auto one.

With the code as it is we will get loads of new BPs called -CE-DH-Warlock-xyz. 
If that's what you want we need to delete all existing BPs before the export.

PR #1079, opened 2022-04-09
Standardize monster blueprint names

ThyWoof:
- also removed 6 monsters spell duplicates (and 2 related conditions) and pointed to SRD Spells instead
- fixed blueprints to GuiNoContent following game guidelines
- fixed some additional missing GUI texts in other areas
- will fix non EN translation files later in the weekend


PhilPJL (approved):

DubhHerder:
In #947 I explained how changing the names would break the monster ai, these changes are exactly what I said should be avoided

PR #1081, opened 2022-04-09
Remove monsters

ThyWoof:


PhilPJL (approved):

DubhHerder:
> Done with complains

What complaints? I haven't said anything since I stepped back before today and today I just warned the authors the monsters no longer worked, which they would have told you themselves as soon as they tested the monsters. I just saved people any wasted time. 

If you had listened to the concerns that I shared over the last few months, this whole issue would have been avoided. It would also have been avoided if you had done literally nothing and left the code as is. 

I explained in pr #947  how changing the names would break the AI , and that's exactly what happened in your standarization PR. Instead of listening ye called my accurate predictions silly. I think that doesn't hold up considering the monsters ai is now broken in the way I said it would be broken. 

Instead of removing a feature from a community mod that others want to use because you are annoyed you could just revert the changes to the functional code and let people use the monsters as they want.

PR #1084, opened 2022-04-09
Refactor Circle of Forest Guardian

PhilPJL:
3 guids changed.


ThyWoof (approved):
only guid changes which is ok for backward compatibility

PR #1111, opened 2022-04-15
fix alternate races removal logic

ThyWoof:


ThyWoof:
forcing merge to fix day zero bugs but I need reviewers...

PR #1113, opened 2022-04-15
fix single class Warlock pact magic

ThyWoof:
- using a constant now


ThyWoof:
another fix without reviewers...

PR #1125, opened 2022-04-15
add Bolgrif race

ThyWoof:


ThyWoof (commented):
moving bug fixes without reviewers

PR #1174, opened 2022-04-19
Character action context

ThyWoof:
- Removed 2 useless patches after I debugged all different sequences during combat
- Added CharacterActionContext with some very useful delegates to get a hold on the executing action or changing some magical effects

These might be useful to make Warlock blasts work correctly as we shouldn't use ancestry type DamageForm for the additional effects.

We could setup a current spell state by setting up a delegate on CastSpell and either use the other delegates which I'm still learning where in code they get invocated (names seem pretty good as an initial tip) or the 3 new interfaces that allow us to have more control over the combat sequence.


ThyWoof:
> Even with those 2 fixes mentioned above it still won't load to Main menu - looks like `CharacterActionContext.Load()` is called before `IGameLocationActionService` service is created

Yep, I was still testing that one and need to probably patch after the location is loaded to register them. I removed this for now and will create another request with the idea.

PR #1192, opened 2022-04-22
Spells display fix

Silvergriffon:
I noticed some custom spells missing from the Spells Display. It turns out that there was a "MinimumSpells" dictionary that is no longer fully implemented. Spells were divided between that dictionary and the SuggestedSpells dictionary by the "suggestedStartsAt" variable in RegisterSpell(). But since the MinimumSpells dictionary isn't used for the display anymore those spells were not being listed. I removed the MinimumSpells dictionary and the suggestedStartsAt variable and now all the spells are available. For example, attached is a before and after of the Warlock spell list
WarlockBeforeAndAfter
.

### DIFF
--- a/SolastaCommunityExpansion/Models/SpellsContext.cs
+++ b/SolastaCommunityExpansion/Models/SpellsContext.cs
@@ -14,14 +14,12 @@ namespace SolastaCommunityExpansion.Models
             private List<string> SelectedSpells => Main.Settings.SpellListSpellEnabled[SpellList.Name];
             public SpellListDefinition SpellList { get; private set; }
             public HashSet<SpellDefinition> AllSpells { get; private set; }
-            //public HashSet<SpellDefinition> MinimumSpells { get; private set; }
             public HashSet<SpellDefinition> SuggestedSpells { get; private set; }
 
             public SpellListContext(SpellListDefinition spellListDefinition)
             {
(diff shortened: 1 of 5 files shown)


ThyWoof:
SuggestedStartedAt is required as these spells are default ones on these lists. Without this we break the classes.

the minimumSpell collection is indeed old code that I forgot to remove.

ThyWoof:
I run the mod in debug mode and dump the CE diagnostics to confirm changes like this won’t break other areas in the mod.

Silvergriffon:
I think something is still not right then. I just misunderstood the cause. None of the spells show up in a spell list if they are before the suggestedStartsAt value. 
For example, Eldritch Blast:
RegisterSpell(EldritchBlast, 1, WarlockSpellList);
It doesn't show up in the Spells Display panel in CE under Warlock. It is not selectable on a Warlock in game. It shows up on other spell lists as a selectable option.

Eldritch Orb:
RegisterSpell(EldritchOrb, 1, WitchSpellList, WarlockSpellList);
Likewise, it doesn't show up in the Spells Display panel in CE under Witch and is not selectable for witches in game, but is selectable on other spell lists.

Basically, anything where the author tried to specify a given spell list that should always get the spell is having the opposite effect: it isn't selectable on that spell list at all.

Issue #1185, opened 2022-04-21
Monsters are broken : Monsters not present in 1.3.44.10, 1.3.44.11, 1.3.46.0


DubhHerder:
no monsters and no monster dll sidercar in release. This would ruin any dungeon that used them. The sidecar idea is obviously unreliable and should be abandoned because of this major flaw 

This is obviously a bug rather than deliberate decision as removing content unilaterally and without notification (or update in changelog) because of bug reports would go against the philosophy of the mod as a "community effort where everything is optional".

The files in the side car do not match the files from PR #947 and the sidecar in general should be removed. Instead PR#947 should be reinstated

ThyWoof:
fixed on https://github.com/SolastaMods/SolastaCommunityExpansion/pull/1195

Issue #1166, opened 2022-04-18
Monsters are broken: high level casters arent using their high level spells


DubhHerder:
The files in the side car do not match the files from PR #947

ThyWoof:
fixed on https://github.com/SolastaMods/SolastaCommunityExpansion/pull/1195

PR #1195, opened 2022-04-22
 fix issues #1185 and #1166

DubhHerder:
took action to fix issues #1185  and #1166


ThyWoof:
Thanks for the fixes but please revert back the sidecar changes. There is a good reason why both Monsters and Multiclass are sidecars. As soon as the monsters fixes are in the next nexus update will have the monsters DLL in the zip. If not enabled, like multiclass, it'll be fully isolated from the rest of CE. Monsters as it's own project will also give you the freedom to work on it independently as I do with Multiclass.

DubhHerder:
Sidecar is too unstable and unreliable , it was accidentally forgotten from the last three releases. Nobody else uses them and I have heard of no good or valid reasons to put this code in a sidecar. This code is finished and complete and does not require constant tweaking like multiclass does. It's not comparable to multiclass, is comparable to feats or spells which remain pay of the primary dll

ThyWoof:
I’m not aware about any instabilities in sidecar code. Nexus releases and builds are separate topics. It wasn’t on Nexus waiting for these fixes. I’ll merge and have the DLL back on next Nexus release once sidecar changes get reversed.

DubhHerder:
> I’m not aware about any instabilities in sidecar code. Nexus releases and builds are separate topics. It wasn’t on Nexus waiting for these fixes. I’ll merge and have the DLL back on next Nexus release once sidecar changes get reversed.

I assumed it was a mistake because I didn't think anyone would remove content without notice. That's a very poor way to handle a community mod.  You mean you purposely removed content without informing any other contributors or users or including the change in the change log?

Why was this the only content removed when a bug was reported, when there have been dozens of other bug reports for other features/codes that have not been removed? 

If content is removed when waiting on fix, warlock should be removed. This does not look like fair treatment.

ThyWoof:
this code throws lots of warnings reason why. merged your files under the sidecar.

PR #1230, opened 2022-05-07
Fixes for various issues

EnderWiggin:
moved patch that allows selecting/deselecting available classes without restart to the MC patch, to make sure it is not called after and overwrites filtered classes with all classes on levelup


EnderWiggin:
Also made spell cast reactions remove lower level options only for not-multiclassed *warlocks*. Previously it removed for all not-multiclassed characters. Also improved default selected slot in case slots options got removed.

PR #1263, opened 2022-05-12
Morwennas spellbook

Silvergriffon:
3 new spells. Illuminating Sphere is an AoE version of Sparkle. Mule makes your party members better at carrying your stuff. Radiant Motes deals tiny damage to many opponents and may make them shed light for a time.


ThyWoof:
@Silvergriffon , you probably created diagnostics with some settings enabled reason why there is a huge list of changed files. Can you please submit a new request without any diagnostics? I usually check them myself before accepting. Thanks! Will close this one. Nice spells by the way.

PR #1278, opened 2022-05-15
Fixes `Shield of Faith` being available to cast blocking reaction cast of `Shield`

EnderWiggin:


EnderWiggin:
Code that searches for a spell that can prevent attack damage takes first spell that can raise AC enough to not get hit, and only afte that checks if that spell has cast time of `Reaction`. Which leads to cases where if `Shield of Faith` is available to cast and is enough to prevent attack (attack roll matches AC or exceeds it by 1) then it will be chosen instead of `Shield` and then discarded because it is not a reaction spell. WHich leaves character without ability to defend himself.

ThyWoof (approved):

PR #1338, opened 2022-06-03
Cleanup text

Vylantze:
Added more cleanups.


ThyWoof:
Awesome, thanks. Keep in mind anything under diagnostics is auto generated and only used for tests. All translations that get into the mod are under Resources.

ThyWoof (approved):

PR #1366, opened 2022-06-06
Class enhancements

Vylantze:
Updated the spell lists of the community expansion classes. Added new infusions to Tinkerer (referencing the original source).

### DIFF
--- a/SolastaCommunityExpansion/Classes/Tinkerer/TinkererClass.cs
+++ b/SolastaCommunityExpansion/Classes/Tinkerer/TinkererClass.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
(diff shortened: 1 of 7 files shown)


ThyWoof:
I like the changes but you are touching 3 different classes from 3 different authors here. Might be easier if you consider a request per class. Discussed this with QA team and other developers and we pretty much agree on Tinkerer. Warlock and Witch are the ones still under discussion.

ThyWoof:
Also keep in mind Witch author will finally be able to put time on this Class again. So it might be better to touch Witch later in the game. We are open for suggestions though.

ThyWoof:
Feedback from @RealBazou : Basically, default/standard Witch spell list should add Dominate Monster, Feeblemind, and Weird.

Vylantze:
Got it, I'll split the changes.

PR #1391, opened 2022-06-13
First draft of magus class

dolenam317:
@ThyWoof character cannot learn spell when level up. Changing spellKnowledge and spellReadyness doesn't help so help is needed here.


ThyWoof:
Will merge and work on it. Will most likely change your patch to follow the same pattern on other battle events. A delegate we can use to make it generic.

ThyWoof (approved):

PR #1396, opened 2022-06-14
Darkelves

Silvergriffon:
Darkelves (Drow)

Some notes:
The sprite is from in-game character images. 

I ran into an issue where feature unlocks granted to races after first level do not properly apply like they do for classes. So, I moved Faerie Fire and Darkness to first level (RAW they are at 3rd and 5th) and reduced the duration of the Darkness effect to be in line with a first level effect.

I added a MorphotypeElementDefinitionBuilder to make a new skin color and hair color.

I changed preferedSkinColors on Bolgrif to not bump into the new skin color for Darkelves.


ThyWoof:
@Silvergriffon , yep, level up panel doesn’t enumerate race progression.

I believe we can fix that later. @EnderWiggin , how hard?

EnderWiggin:
Need to look closer. Sounds easy-ish at this moment, but has potential for some complications.

ThyWoof (approved):

ThyWoof:
Will merge this and rollback some stuff after.

1) Ensure all new translations reside on their proper files;
2) Translations.zip shouldn't be commit. Probably your .gitignore wasn't up to date;

PR #1433, opened 2022-06-19
Change Soulblade's Soul Empower to Free Action

FelipeRenault:
## The problem

- Soul Empower is a feature designed to be used right after a long rest. However, while travelling, it is not possible to use it right after each long rest, making your first turn on a road combat be subpar, which is even more problematic when your party is surprised or asleep

## The solution

- Chaging Soul Empower's cost to a Free Action is a nice QoL, since you're expected to use it right after each long rest anyway, so the Action cost wasn't usually meaningful, except on the scenario described above


ThyWoof:
Mage Armor suffer from the same as well. The ideal solution is to find a way to patch the world map. Spoke with the team and they prefer to keep as is.

PR #1563, opened 2022-07-18
Fix issues with spell strike and magus

dolenam317:
- Only cantrip that have attack roll is qualified for spell strike.
- Don't skip the spell if the attack from spell strike miss because it will cause mod instability warning.
- Fix Spell Strike show as None in combat log.

@EnderWiggin  thank you for pointing out those issues.


EnderWiggin (approved):
Looks fine. Although I'd not remove non-attack cantrips from being eligible for Spell Strike. But that can be a separate discussion and part of a different improvement.

ThyWoof (approved):

PR #1729, opened 2022-11-21
Add druid subclass Circle of the Night

tivie:
Add the subclass Circle of the Night, inspired by the official Circle of the Moon subclass.

### 3rd level
 - **Combat Wild Shape**: Lowers the level requirements for stronger beasts and adds a new beast called Dire Bear, inspired by the Cave Bear monster.
 - **Combat Healing**: While wild-shaped, you can use a bonus action to heal yourself for 1d8 hit points. You can use this feature a number of times equal to your proficiency modifier per form you adopt per long rest.

### 6th level
 - **Primal Strike**: Starting at 6th level, your attacks in beast form count as magical for the purpose of overcoming resistance and immunity to nonmagical attacks and damage
 - **Improved Combat Healing**: Your Combat Healing now heals you for 2d8+3 hit points

### 10th level
 - **Elemental Form**:
 - **Superior Combat Healing**: Your Combat Healing now heals you for 2d8+3 hit points

## Minor Issues:

 - Combat Healing also allows for healing in "human form", which it shouldn't.
 - Combat Wild Shape doesn't have any referencing in the subclass description screen when selecting the subclass. I couldn't add the feature to the subclass selection screen without messing with the in-game UI.
 - Elemental Form only expects one wild shape charge instead of two. But the elemental forms are nerfed compared to the TT version, so it's balanced.
 - Elemental Form doesn't have any reference in the class description screen.

### DIFF
--- a/SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
+++ b/SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
@@ -1,0 +1,339 @@
+﻿using SolastaUnfinishedBusiness.Builders;
+using SolastaUnfinishedBusiness.Builders.Features;
+using static RuleDefinitions;
+using static SolastaUnfinishedBusiness.Api.DatabaseHelper;
+using static SolastaUnfinishedBusiness.Api.DatabaseHelper.CharacterSubclassDefinitions;
+using static SolastaUnfinishedBusiness.Api.DatabaseHelper.FeatureDefinitionPowers;
+using static SolastaUnfinishedBusiness.Api.DatabaseHelper.MonsterDefinitions;
+using static SolastaUnfinishedBusiness.Api.DatabaseHelper.MonsterAttackDefinitions;
+using static EffectForm;
+using SolastaUnfinishedBusiness.Patches;
+
+namespace SolastaUnfinishedBusiness.Subclasses;
+
+internal sealed class CircleOfTheNight : AbstractSubclass
+{
+    private const string CircleOfTheNightName = "CircleOfTheNight";
+
+
+    // custom wildshapes
+
+    /**
+     * based on MM Cave Bear
+     * */
+    internal static MonsterDefinition HBWildShapeDireBear()
+    {
+        // attacks
+        // Bite
+        // TODO Bump damage mod from +4 to +5
+        var biteAttack = new MonsterAttackIteration
+        {
+            monsterAttackDefinition = MonsterAttackDefinitionBuilder
+                .Create(Attack_Wildshape_BrownBear_Bite, "Attack_Wildshape_DireBear_Bite")
+                .SetToHitBonus(7)
+                .AddToDB()
+        };
+
+        // Claw
+        var clawAttack = new MonsterAttackIteration
+        {
+            monsterAttackDefinition = MonsterAttackDefinitionBuilder
+            .Create(Attack_Wildshape_BrownBear_Claw, "Attack_Wildshape_DireBear_Claw")
+            .SetToHitBonus(7)
+            .AddToDB()
+        };
+
+        MonsterDefinition shape = MonsterDefinitionBuilder.Create(WildshapeBlackBear, "WildShapeDireBear")
+            // STR, DEX, CON, INT, WIS, CHA
+            .SetAbilityScores(new int[] { 20, 10, 16, 2, 13, 7 })
+            .SetArmorClass(14)
+            .SetStandardHitPoints(42)
+            .SetHitDice(DieType.D10, 5)
+            .SetChallengeRating(2)
+            .SetOrUpdateGuiPresentation("Dire Bear", "A really big bear", WildshapeBlackBear)
+            .SetAttackIterations(new MonsterAttackIteration[] { biteAttack, clawAttack })
+            .AddToDB();
+
+        return shape;
+    }
+
+    internal static MonsterDefinition HBWildShapeAirElemental()
+    {
+        MonsterDefinition shape = MonsterDefinitionBuilder.Create(Air_Elemental, "WildShapeAirElemental")
+            // STR, DEX, CON, INT, WIS, CHA
+            .SetAbilityScores(new int[] { 14, 20, 14, 6, 10, 6 })
+            .SetArmorClass(15)
+            .SetStandardHitPoints(90)
+            .SetHitDice(DieType.D10, 12)
+            .AddToDB();
+
+        return shape;
+    }
+
+    internal static MonsterDefinition HBWildShapeFireElemental()
+    {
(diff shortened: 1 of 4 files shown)


tivie:
Related: #1726

SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
```diff
@@ -0,0 +1,339 @@
...
+
+    internal CircleOfTheNight()
+    {
+        // 3rd level
+        // Combat Wildshape 
+        // Official dnd 5e rules are CR = 1/3 of druid level. However in solasta the selection of beasts is greatly reduced
+        var combatWildshape = FeatureDefinitionPowerSharedPoolBuilder
+            .Create("PowerDruidCombatWildShape")
+            .SetOverriddenPower(PowerDruidWildShape)
+            .SetGuiPresentation(Category.Feature, hidden: true)
```
EnderWiggin, line 268: this power is set to have hidden representation - this is the cause of it not being shown in the subclass features

SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
```diff
@@ -0,0 +1,339 @@
...
+            .SetSharedPool(ActivationTime.BonusAction, PowerDruidWildShape)
+            .SetEffectDescription(BuildCombatWildShapeEffectDescription())
+            .AddToDB();
+
+        // Combat Wild Shape Healing
+        // While wildshaped, you can use a bonus action to heal yourself for 1d8 hit points.
+        // You can use this feature a number of times equal to your Wisdom modifier per long rest
+        var combatHealing = FeatureDefinitionPowerBuilder
```
EnderWiggin, line 276: you can use `ValidatorsPowerUse` sub-feature to make this power hidden when not-wild shaped

SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
```diff
@@ -0,0 +1,339 @@
...
+        };
+
+        MonsterDefinition shape = MonsterDefinitionBuilder.Create(WildshapeBlackBear, "WildShapeDireBear")
+            // STR, DEX, CON, INT, WIS, CHA
+            .SetAbilityScores(new int[] { 20, 10, 16, 2, 13, 7 })
+            .SetArmorClass(14)
+            .SetStandardHitPoints(42)
+            .SetHitDice(DieType.D10, 5)
+            .SetChallengeRating(2)
+            .SetOrUpdateGuiPresentation("Dire Bear", "A really big bear", WildshapeBlackBear)
```
EnderWiggin, line 53: do not hardcode translation strings

SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    internal static EffectDescription BuildCombatWildShapeEffectDescription()
+    {
+        var WildShapeEffect = EffectDescriptionBuilder
+            .Create()
+            .SetEffectAdvancement(EffectIncrementMethod.None)
+            .Build();
+
+        //WildShapeEffect.targetParameter = 1;
```
EnderWiggin, line 114: Better to use EffectDescriptionBuilder for setting up this effect

ThyWoof (changes requested):
no need to touch DatabaseHelper-RELEASE. I manually maintain this file on my end. You can do your dev using the DEBUG build instead which has all blueprints definitions.

SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
tivie: Yes. Wild shape has a special bonus action button in the UI. When combatWildshape is not hidden, a duplicate appears in the "Powers" menu of the combat UI. The behavior of that duplicated button is super buggy (if you click it while wild shaped, it expends your bonus action and a wild shape charge, but has no effect).

SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
tivie: fixed

SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
ThyWoof: In this case you should use a FeatureSetBuilder just for the UI part. Add above power as a FeatureSet and you'll have the expected results. Check the feature sets I create on Wild Master exactly because of this reason.

SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
ThyWoof: here is an example on how to validate powers. https://github.com/SolastaMods/SolastaUnfinishedBusiness/blob/ec9a1288bac4373af3d77d3763ed135ffdd00593/SolastaUnfinishedBusiness/Subclasses/WayOfTheSilhouette.cs#L58

SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
ThyWoof: Also, I don't believe you need to use a shared pool here. Might be simpler to replace the PowerWildShape that is granted on level 2 with the enhanced one you create.

ThyWoof (approved):
I'll approve as I believe the power shared pool can be much simplified and would like to test an idea. We still need though to clean up the EffectDescription creation in favor of our builders.

SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
tivie: It seems that with the changes, the Wild Shape button is buggy:

 - When you Wild Shape, it doesn't change the consumed charges in the button. But wild shaping two times makes the button disappear.
 - The wild shape button appears even when wild shaped. Clicking it and selecting a new form turns you momentarily into a new form, which drops immediately. It doesn't consume any charge though.


I will work on this, using your suggestion of using `FeatureDefinitionFeatureSetBuilder` to display the ability in the character creation screen.

ThyWoof: Keep in mind Wildshape and Pools that use WildShape as a shared power have a crazy interaction. I had some issues in the past reason why was trying to fully replace the WildShape power with your enhanced version. Try going on this path first.

tivie: I see. I suspect the issue might be with the special Wild Shape button. Though I dont know if its easy to create a new one from scratch to replace the existing one.

PR #1744, opened 2022-11-23
Circle of the Night bug fixes and Water Elemental

tivie:
Circle of the Night:

 - Fix bug that makes wild shape form drop after a few seconds
 - Implement Water Elemental based off Air Elemental sprite.


tivie:
Seems it was already fixed. Ignore this!

PR #1760, opened 2022-11-25
Torchbearer as a Fighting Style

ElAntonius:
This moves torchbearer to a fighting style. This fighting style adds a bonus torch attack, which automatically hits and then allows a save for the on fire condition.

image

Also, bundled in some minor fixes: added a Silent to the DualFlurryApply feature so the "X has dual flurry' combat log message only appears when the character gains the bonus action, and cleaned up a typo on my name in the credits.

image


SolastaUnfinishedBusiness/Feats/TwoWeaponCombatFeats.cs
```diff
@@ -37,6 +37,7 @@ private static FeatDefinition BuildDualFlurry()
             .SetPossessive()
             .SetConditionType(ConditionType.Beneficial)
             .SetSpecialInterruptions(ConditionInterruption.AnyBattleTurnEnd)
+            .SetSilent(Silent.WhenAddedOrRemoved)
```
ThyWoof, line 40: whenever silent we don't really need to setup a GUI presentation and use NoContent instead. will merge and make this change after.

ThyWoof (approved):

PR #1784, opened 2022-11-28
PathOfTheSpirits: fix Wolf Leadership

### LINKED ISSUE #1783: Path of the Spirits barbarian: Wolf Leadership gives advantage even after rage end
After the last update, Wolf Leadership is giving advantage even after the barbarian stops raging.

Steps to reproduce: 

1. Engage battle
2. Attack (no advantage to self or friends)
3. Barbarian Rages
4. Attack (advantage)
5. Barbarian Stops Raging
6.  Attack (advantage) --> should not have advantage since the barbarian stopped raging.

tivie:
Now, Wolf leadership only gives advantage if the barbarian is raging.

Closes #1783

### DIFF
--- a/SolastaUnfinishedBusiness/Subclasses/PathOfTheSpirits.cs
+++ b/SolastaUnfinishedBusiness/Subclasses/PathOfTheSpirits.cs
@@ -205,11 +205,11 @@ internal sealed class PathOfTheSpirits : AbstractSubclass
             .SetCustomSubFeatures(PowerVisibilityModifier.Hidden)
             .SetUsesFixed(ActivationTime.OnRageStartAutomatic)
             .SetEffectDescription(EffectDescriptionBuilder.Create()
-                .SetTargetingData(Side.Ally, RangeType.Self, 0, TargetType.Sphere, 6)
+                .SetTargetingData(Side.Ally, RangeType.Self, 1, TargetType.Sphere, 6)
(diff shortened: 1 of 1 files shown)


ThyWoof:
Yesterday I changed Wolf to have same triggers as Bear after some non-working reports on Nexus. Are you sure it doesn't work as is? Both bear and wolf have a interrupt trigger on rage stop

SolastaUnfinishedBusiness/Subclasses/PathOfTheSpirits.cs
```diff
@@ -205,11 +205,11 @@ private static FeatureDefinition PowerPathOfTheSpiritsWolfLeadership()
             .SetCustomSubFeatures(PowerVisibilityModifier.Hidden)
             .SetUsesFixed(ActivationTime.OnRageStartAutomatic)
             .SetEffectDescription(EffectDescriptionBuilder.Create()
-                .SetTargetingData(Side.Ally, RangeType.Self, 0, TargetType.Sphere, 6)
+                .SetTargetingData(Side.Ally, RangeType.Self, 1, TargetType.Sphere, 6)
```
ThyWoof, line 208: You can keep range type 0 here…

ThyWoof (changes requested):

ThyWoof (approved):

PR #1785, opened 2022-11-28
Add fs for harlequin

dolenam317:


dolenam317:
@ThyWoof @EnderWiggin could you please take a look. This is just a draft PR but if `FeatureDefinitionFightingStyleChoice` is unlocked at level 3 then level up pannel show nothing. Moving to level 4 and then things are fine.

ThyWoof (approved):

PR #1818, opened 2022-12-03
Update WayOfTheSilhouette-en.txt

sztrzask:
Updated text. Looking at the code [CustomConditionsContext.cs](https://github.com/SolastaMods/SolastaUnfinishedBusiness/blob/master/SolastaUnfinishedBusiness/Models/CustomConditionsContext.cs)  I'm not sure if moving OUT of darkness would break the invisibility, but it looks like performing any action outside the darkness would (except teleport and self-bufs).


ThyWoof:
It should behave exactly like Moonlit. Would you please test that one as well? If so we need to tweak text there

sztrzask:
Sure, I'll do that tomorrow

ThyWoof (approved):

sztrzask:
@ThyWoof they both seem to work as writtern - including "stepping into the light removes the invisibility" (that one I wasn't sure about)

PR #1928, opened 2022-12-13
Tweaks

EnderWiggin:


EnderWiggin:
Currently there's small issue that characters with Planar Magic feats will see Invocations button on the panel, even if they don't know any regular invocations. Clicking it shows empty panel. Will try to fix a bit later.

EnderWiggin:
- fixed Plane Magic button not being visible if no spells from it can be cast (e.g. hands are full)
- fixed positioning of power use slots in invocation box under different screen resolutions
- fixed Fey Teleportation's Misty Step not requiring short rest to recharge
- made custom spell granting invocations(Plane Magic) show custom sub-title in tooltip, instead of default Eldritch Invocation

ThyWoof (approved):

PR #1992, opened 2022-12-19
Tiefling tweaks

Silvergriffon:
I noticed that tiefling had no horns unless I enabled horns for all races. Turns out it wasn't inheriting from the main race. I fixed that, added infernal names and set preferred skin colors. You probably have a better way to fix it. The male horns are a real issue though. They really don't fit properly at all. Maybe solasta's tieflings only have horny females??


ThyWoof:
Have you tried female horns on males?

ThyWoof (approved):

Silvergriffon:
> Have you tried female horns on males?

They use the same blueprints and the game renders the version of the 3d model appropriate to the gender. Horns_A is a different model for males vs females. I don't see a way to trick it into using the other model.

PR #2259, opened 2023-01-28
Quick fix to soulblade's weapon empower not limiting 2h weapons

Trokkin:
Self-explanatory. Attached a CustomItemFilter to the empowering feature, so that it would filter out non-proficient or 2h weapons.

~~Btw, is there a discord server or smth for more informal questions about the mod codebase? I'm trying to recreate Monk's dedicated weapon feature, which in simplified version is very similar to bladelock's empowering, but somehow `AbilityScoreReplacement.DexterityIfBetterThanStrength` isn't working, so I had to did `.SetCustomSubFeatures(new CanUseAttributeForWeapon` thing.~~ Edit: nevermind, i found the discord button in the mod. im blind lol

### DIFF
--- a/SolastaUnfinishedBusiness/Subclasses/PatronSoulBlade.cs
+++ b/SolastaUnfinishedBusiness/Subclasses/PatronSoulBlade.cs
@@ -1,4 +1,6 @@
-﻿using SolastaUnfinishedBusiness.Builders;
+﻿using System.Linq;
+using SolastaUnfinishedBusiness.Api.Extensions;
+using SolastaUnfinishedBusiness.Builders;
 using SolastaUnfinishedBusiness.Builders.Features;
 using SolastaUnfinishedBusiness.CustomBehaviors;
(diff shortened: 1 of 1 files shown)


kosocku:
Question about the Soulblade empowered weapon thing: Improved Pact Weapon invocation removes that restriction in 5e, so will you be able to implement that into the current invocation?  If not, aren't you better off leaving it as-is until that part can be figured out?

Trokkin:
I think checking if character has the Improwed Pact Weapon feature should suffice, I'll check if it works

Trokkin:
So that was a success, now 2-handed melee can be selected with blade pact and then ranged with the improving invocation.

I couldn't find how to check the pact itself, since it's a feature Set, not a feature in itself. Also, the validation function is now a bit bloated but pretty straight-forward. So, I welcome any refactor to that function, but at least it works

ThyWoof (approved):

PR #2283, opened 2023-01-30
Skin of Retribution

Pikachar2:
Added in 'Skin of Retribution'
- 1st Level Spell (Normally for Warlocks)
- School: Abjuration
- Range: Self
- Duration: 1 hour
- Gives Temporary HP and a retributive (currently Cold) damage on melee attacks that both scale with spell level (5 hp/damage per spell level).
- If caster runs out of TempHP they lose the retribution damage.


ThyWoof:
Curious on what the original spell name is.

ThyWoof (approved):

PR #2304, opened 2023-02-02
Skin of Retribution Upgrade

LunaKB:
You can now change the damage type of Skin of Retribution.


Pikachar2:
Note: These changes were made to make this spell distinct from the PHB/NON-SRD version of the spell: Armor of Agathys. Previously it was just a name change. This makes it so that it is a unique spell that does not "really" increase it's power enough (especially in Solasta) to increase it in spell level.
It also leaves the flavor intact, so that if a purist wants the original, they can just select "Cold" damage. Though, the other damage types are still in the same "group" of damage types. It's not like it got changed to radiant or force damage.
The versatility alone is not enough to increase the spell level and it is functionally the same.

ThyWoof (approved):

PR #2356, opened 2023-02-04
WIP on Spear Mastery

EnderWiggin:


EnderWiggin:
Made Spear mastery's  AoO properly trigger for spears only and bonus damage is applied only to that triggered attack.

ThyWoof (approved):

PR #2400, opened 2023-02-09
Tweaks and fixes

EnderWiggin:
Made mod settings checkboxes use icons Also made bunch of improvements to Fell Handed :

- Log advantage prone trigger even when it misses + properly show hit trends
- properly account for crit hits/misses and difficulty modifiers when deciding if it hit
- add separate line saying that target is prone on success
- fixed disadvantage miss damage ignoring damage affinities ( immunity/resistance/vulnerability)
- added log tat feature is triggered to damage on disadvantage miss


EnderWiggin:
Icons probably need shrinking from 16x16 to 12x12

ThyWoof (approved):

PR #2508, opened 2023-02-19
fixed war magic to allow for attacks with ranged weapons.

BGrym:


SolastaUnfinishedBusiness/Subclasses/_CommonBuilders.cs
```diff
@@ -82,7 +83,13 @@ internal static class CommonBuilders
                 .SetConditionForm(ConditionDefinitionBuilder
                         .Create("ConditionCasterFightingWarMagic")
                         .SetGuiPresentationNoContent(true)
-                        .AddFeatures(DatabaseHelper.FeatureDefinitionAttackModifiers.AttackModifierBerserkerFrenzy)
+                        .AddFeatures(FeatureDefinitionAttackModifierBuilder
+                            .Create($"PowerCasterFightingWarMagicAttack")
+                            .SetGuiPresentation("PowerCasterFightingWarMagicAttack", Category.Feature)
```
ThyWoof, line 88: We need to pass here the string to `PowerCasterFightingWarMagic` instead of `PowerCasterFightingWarMagicAttack`.

SolastaUnfinishedBusiness/CustomBehaviors/AddExtraAttack.cs
```diff
@@ -230,7 +229,7 @@ protected override List<RulesetAttackMode> GetAttackModes([NotNull] RulesetChara
         var mainHandItem = hero.CharacterInventory.InventorySlotsByName[EquipmentDefinitions.SlotTypeMainHand]
             .EquipedItem;
 
-        var strikeDefinition = mainHandItem.ItemDefinition;
+        var strikeDefinition = mainHandItem?.ItemDefinition ?? hero.UnarmedStrikeDefinition;
```
ThyWoof, line 232: what is the use case here?

ThyWoof (changes requested):

SolastaUnfinishedBusiness/CustomBehaviors/AddExtraAttack.cs
BGrym: Not having the null check was causing issues if the main hand was empty. This should allow an unarmed strike from the bonus action in that case, but I could also just return on null instead if it's preferred to not allow for the attack if the main hand is empty.

SolastaUnfinishedBusiness/Subclasses/_CommonBuilders.cs
BGrym: `PowerCasterFightingWarMagic` is the name of the `FeatureDefinitionPower`, but since I'm adding a `FeatureDefinitionAttackModifier` to it I thought the attack feature needed a different name. I'll change it if that's not the case though.

ThyWoof: Idea is to minimize number of terms we need to create. First parameter there will bypass the default logic that takes the BP name to form the gui. We already have a power that has the terms we need to this attack modifier. Reason for the change.

BGrym: Thanks for the explanation! Fixed.

ThyWoof (changes requested):
Will wait for the additional comment on AddExtraAttack and the fix on Gui on the new created power to merge.

ThyWoof (approved):

PR #2604, opened 2023-02-26
Update MeleeCombatFeats.cs

Tintelpe:
Check the battle variable to make sure the ref exists; handles a bug where using a weapon outside of combat causes a null deref.

I don't just early return because this will allow blademaster rogues to apply the feat on a back stab; the validator should still pass and battle is only used in the early exit for the special case -- I'm not familiar enough with the codebase to understand precisely what that case is handling.


ThyWoof (approved):

ThyWoof:
correct. thanks for the fix.

PR #2653, opened 2023-03-05
Updated descriptions Royal Knight Fighter subclass abilities

LunaKB:
- Changed descriptions for the following Royal Knight Fighter subclass abilities:
-- Rallying Cry
-- Inspiring Surge
- Corrected title for Inspiring Surge power. It used to say "Preserve Life". It now says "Inspiring Surge".

This should provide clarity to the players.


SolastaUnfinishedBusiness/Subclasses/MartialRoyalKnight.cs
```diff
@@ -31,7 +31,7 @@ internal MartialRoyalKnight()
             .AddToDB();
 
         var effectDescription = EffectDescriptionBuilder
-            .Create(PowerDomainLifePreserveLife.EffectDescription)
+            .Create()
```
ThyWoof, line 34: Why are you touching these effects?

ThyWoof (changes requested):

SolastaUnfinishedBusiness/Subclasses/MartialRoyalKnight.cs
```diff
@@ -57,11 +57,11 @@ internal MartialRoyalKnight()
             .AddToDB();
 
         var powerRoyalKnightInspiringSurge = FeatureDefinitionPowerBuilder
-            .Create(PowerDomainLifePreserveLife, "PowerRoyalKnightInspiringSurge")
```
ThyWoof, line 60: Why change this power?

SolastaUnfinishedBusiness/Subclasses/MartialRoyalKnight.cs
```diff
@@ -57,11 +57,11 @@ internal MartialRoyalKnight()
...
+            .Create("PowerRoyalKnightInspiringSurge")
             .SetGuiPresentation(Category.Feature, SpellDefinitions.Heroism)
             .SetUsesFixed(ActivationTime.BonusAction, RechargeRate.LongRest)
             .SetEffectDescription(EffectDescriptionBuilder
-                .Create(PowerDomainLifePreserveLife.EffectDescription)
```
ThyWoof, line 64: Why change this effect?

ThyWoof (changes requested):

SolastaUnfinishedBusiness/Subclasses/MartialRoyalKnight.cs
Pikachar2: The ability effectively gives Action Surge to another character. It has nothing to do with Preserve Life. Which is the Life Cleric's Channel Divinity. To the best of our knowledge it seems that it's just a description/title that's setup as kind of a template in these instances. And on top of that, it was named as Preserve Life when looking for the ability in Combat/from the Action bar.

ThyWoof: These have nothing to do with descriptions but the template we copy behavior from. I don’t believe it should be changed. Only the translations

SolastaUnfinishedBusiness/Subclasses/MartialRoyalKnight.cs
LunaKB: When you select the power as a bonus action, the name for power was incorrect. Removing this makes the text correct. Here's what the UI looks like without the fix:

UnfinishedBusiness-pull2653

SolastaUnfinishedBusiness/Subclasses/MartialRoyalKnight.cs
LunaKB: This was changed in the effort to get the text for the power Inspiring Surge to no longer say "Preserve Life". I'll update this with this put back.

SolastaUnfinishedBusiness/Subclasses/MartialRoyalKnight.cs
LunaKB: This was changed in the effort to get the text for the power Inspiring Surge to no longer say "Preserve Life". I'll update this with this put back.

LunaKB:
Updated pull request according to comments.

SolastaUnfinishedBusiness/Subclasses/MartialRoyalKnight.cs
ThyWoof: this is an effect description. It's nothing to do with the description on screen.

ThyWoof:
I confirm it’s indeed showing the wrong title even though the json dump says otherwise. Will look into it

LunaKB:
Alright, if it will be fixed later, I'll update with the preserve life put back.

LunaKB:
Updated pull request according to comments.

ThyWoof:
found out what the issue was. PowerDomainLife also had a short title description set. This was causing the wrong title to display.

PR #2910, opened 2023-04-03
Minor fixes to Item Crafting

Berserk-Knight:
Fixed the Punisher Greatsword recipe not removing the Primed Battleaxe from its ingredient list due to a typo.
Made the newly generated recipes use primed items instead of base items when available.
* List will have to be updated manually, should more types of primed items be added.


ThyWoof (approved):

ThyWoof:
thanks

PR #3208, opened 2023-05-13
Added icons for gravity spells

kloranger:


ThyWoof:
Thanks. There is way more changes here than the new icons so I downloaded them 4 and merged.

PR #3213, opened 2023-05-13
add chinese translation of PathOfTheYeoman

magicskysword:
add chinese translation of PathOfTheYeoman


ThyWoof (approved):

ThyWoof:
thanks. realized had a copy and paste issue on subclass title.
should be "Path of the Yeoman". What would be the Chinese term?

magicskysword:
the Chinese translation has been translated from "Path of the Yeoman", and Yeoman has a lot of meanings, I'm not sure which one, but in Chinese, I translated it as "自卫军", meaning self-defence corps.If it means something else, I'll correct it.

ThyWoof:
This sub is all about a Barb being a fletcher. That should be the meaning.

PR #3265, opened 2023-05-24
PATCH: add a rule to grant adv to attacks that the target unable to see

HIEROT:
Tested in the master branch (Because by the time I pulled the dev branch it was under modification hence failed in compiling)
Now you can notice that when I attack the bandit without darkvision, I have advantage. And the adv and disadv cancel if both party don't have darkvision.
image
image

### DIFF
--- a/SolastaUnfinishedBusiness/Patches/GameLocationBattleManagerPatcher.cs
+++ b/SolastaUnfinishedBusiness/Patches/GameLocationBattleManagerPatcher.cs
@@ -15,6 +15,7 @@ using SolastaUnfinishedBusiness.CustomInterfaces;
 using SolastaUnfinishedBusiness.CustomValidators;
 using SolastaUnfinishedBusiness.Models;
 using TA;
+using UnityEngine;
 
 namespace SolastaUnfinishedBusiness.Patches;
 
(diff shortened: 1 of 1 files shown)


HIEROT:
Oh another TODO: add a switch in the setting panel to on/off this modification.

ThyWoof:
Yep, DEV now is on POI version. Can you hold this one 24 hrs and resubmit once we publish 1.5?

ThyWoof (approved):

HIEROT:
Invisibility and this feature will grant two stacking advantages, which doesn't make sense if you calculate the result using the vanilla solasta adv rules because they both cite the same rule (combat with enemies that cannot be seen). But if you switch on official adv rules, it seems doesn't matter. That being said, a better solution may be moving this block of code to somewhere invisibility is handled and merging them, but I just haven't yet identify it.

ThyWoof:
I tried patching RuleSetCharacter.CanSenseTarget to return true of target has greater invisibility and it works fine but condition gets removed. My problem now is to prevent this to happen. Not sure where in code it remove conditions based on failed senses

ThyWoof:
By the way it ended up merged with a setting

HIEROT:
Oh I tested a senario where darkness spell is covered. Currently, two characters without true sight, one insight and the other outside the affected region, will attack each other with disadvantage, which shouldn't be the case. It's true that my modification is far from perfect because only light condition is hard coded here :(. It's not difficult to fix this right at the location I modified, but a better solution may be like what you said, we should check if the two creatures can see each other and apply advantage rules accordingly (in a more elegant way)

PR #3273, opened 2023-05-25
BUGFIX: Fix Darkness Spell interaction with Devil's Sight and similar ability

HIEROT:
Now that we have better handled sight and advantage elsewhere (if switched on), darkness should no more give other source of disadvantage on attack. Besides, the immunity to condition darkness provided by the condition affinity prevents one with devil sight and similar abilities from inflicting condition darkness, causing other problems (like attackers with darkvision ignore your obscurity as if you are in normal unlit condition) so I removed this immunity.

### DIFF
--- a/SolastaUnfinishedBusiness/Api/DatabaseHelper-RELEASE.cs
+++ b/SolastaUnfinishedBusiness/Api/DatabaseHelper-RELEASE.cs
@@ -1,4 +1,4 @@
-// manually generated on 04/10/23
+﻿// manually generated on 04/10/23
 using TA.AI;
 // ReSharper disable IdentifierTypo
 // ReSharper disable InconsistentNaming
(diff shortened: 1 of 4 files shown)


HIEROT:
As in the screenshot, now the darkness functions as expected.
屏幕截图 2023-05-26 025805

ThyWoof (commented):
Can you code in a way that doesn’t require restart? So instead of exiting early if setting not enabled please bring those 2 vanilla blueprints back to their default values. Search for methods Switch* on other UI displays to follow the same design.

HIEROT:
I will take a look

HIEROT:
Done with it)

ThyWoof (approved):
Thanks. Almost there but if you don't call the fix as well during mod load it will force users to go back to mod UI and flip it again. Will fix after merge.

ThyWoof:
this: https://github.com/SolastaMods/SolastaUnfinishedBusiness/commit/1623f5821655e28eb7917244814515deede1611c

PR #3319, opened 2023-05-28
Eliminate allocations

PhilPJL:
Hi Zappa good to see you're keeping up the good work.  
We can't have you using GCHandle.Alloc, so last century.  Try this alloc free approach.  
Requires System.Memory.
I played with this enabled and nothing bad happened, but that's as far as my testing went.


ThyWoof:
Thanks. I felt back in college calling alloc.

ThyWoof (approved):

PR #3390, opened 2023-06-03
Fix abilities of the Chain Master only takes effect in combat

HIEROT:
Using a little trick to check if a character is in battle, now this invocation successfully applies out of combat.

### DIFF
--- a/SolastaUnfinishedBusiness/CustomBuilders/InvocationsBuilders.cs
+++ b/SolastaUnfinishedBusiness/CustomBuilders/InvocationsBuilders.cs
@@ -1,4 +1,4 @@
-using System.Collections;
+﻿using System.Collections;
 using System.Linq;
 using SolastaUnfinishedBusiness.Api.GameExtensions;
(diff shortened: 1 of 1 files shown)


SolastaUnfinishedBusiness/CustomBuilders/InvocationsBuilders.cs
```diff
@@ -616,12 +616,12 @@ private sealed class AfterActionFinishedAbilitiesChain : IActionFinished
 
         public IEnumerator OnActionFinished(CharacterAction action)
         {
-            if (action.ActionType != ActionDefinitions.ActionType.Bonus)
+            if (action.ActingCharacter == null)
```
ThyWoof, line 619: change partner here. ActingCharacter AFAIK will always be non null.

should be:

if (action.ActingCharacter.RulesetCharacter is not {IsDeadOrDyingOrUnconscious: false})

ThyWoof (changes requested):

PR #3395, opened 2023-06-04
Fixed bug where Chinese, Japanese and Korean did not wrap the line correctly

magicskysword:
# Why fix this
https://github.com/SolastaMods/SolastaUnfinishedBusiness/issues/3394
Solasta's way of wrapping text is to use Spaces to break up words, but there are no Spaces between characters in CJK(Chinese, Japanese and Korean), so there is a wrapping error.

# Principle of repair
Changed how to Split CJK text, and eliminates the spacing between CJK character.

# How to enable
This fix is switchable because I am concerned about the possible impact on other languages.(Although according to my tests there should be no other effect)
It's in Interface > Game UI > Campaigns and locations. I don't know where to put it right now.

### DIFF
--- a/SolastaUnfinishedBusiness/Displays/GameUiDisplay.cs
+++ b/SolastaUnfinishedBusiness/Displays/GameUiDisplay.cs
@@ -140,6 +140,13 @@ internal static class GameUiDisplay
                 Main.Settings.DontFollowMargin = intValue;
             }
         }
+        
+        UI.Label();
+        toggle = Main.Settings.FixCJKWrappingError;
+        if (UI.Toggle(Gui.Localize("ModUi/&FixCJKWrappingError"), ref toggle, UI.AutoWidth()))
+        {
+            Main.Settings.FixCJKWrappingError = toggle;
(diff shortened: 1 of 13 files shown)


ThyWoof (approved):

kappa-maintainer:
If you are really concern about wrong wrapping on other languages, you can try using BreakIterator from unicode's official tool icu-dotnet

ThyWoof:
Can you please open a discussion and share more? I don’t speak any Asian language at all and only recently decided support them.

ThyWoof:
https://github.com/sillsdev/icu-dotnet @magicskysword

magicskysword:
> If you are really concern about wrong wrapping on other languages, you can try using BreakIterator from unicode's official tool icu-dotnet

According to my tests, the ICU library is not suitable.It requires specifying the language type and will removes some code symbols incorrectly (such as "<sprite_index={0}>") . The library is also not platform-wide; each platform uses a different build. Just a single platform build is 25mb, which is too large.

PR #3438, opened 2023-06-08
Fix translations of CN

YuanHuaiYu:
Fix wrongly translated VSM component title in Rules-zh-CN.
Make them to use the same word that used in Setting-zh-CN.


magicskysword:
The file in Diagnostics does not need to be modified, that is the reference file.
在Diagnostics里的文件不用改，那个是导出来的参考文件，不参与构建。只改UnofficialTranslation里的就行

ThyWoof:
I suggest for anything Chinese please work on @magicskysword fork. I prefer to have only one point of contact per language.

PR #3444, opened 2023-06-08
Circle of the Night wildshape model scale

Daieumon:
increase model scale of Ape and TundraTiger to fill out 4 tiles. Only applies to circle of the night

### DIFF
--- a/SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
+++ b/SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
@@ -152,10 +152,10 @@ internal sealed class CircleOfTheNight : AbstractSubclass
             ShapeBuilder(2, WildShapeBrownBear),
             ShapeBuilder(4, WildshapeDeepSpider),
             ShapeBuilder(4, HbWildShapeDireBear()),
-            ShapeBuilder(6, WildShapeApe),
+            ShapeBuilder(6, HbWildShapeApe()),
             ShapeBuilder(8, WildshapeTiger_Drake),
             ShapeBuilder(8, WildShapeGiant_Eagle),
-            ShapeBuilder(10, WildShapeTundraTiger),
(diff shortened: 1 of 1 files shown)


ThyWoof (changes requested):
How well will this integrate with the mod setting that make all WS 1 cell in size?

Have you tested this scenario?

Daieumon:
fixed bug with the mod setting.

SolastaUnfinishedBusiness/Subclasses/CircleOfTheNight.cs
```diff
@@ -228,6 +228,42 @@ private static MonsterDefinition HbWildShapeDireBear()
...
+        return shape;
+    }
+
+    private static MonsterDefinition HbWildShapeApe()
+    {
+        var shape = MonsterDefinitionBuilder
+            .Create(WildShapeApe, "WildshapeApe")
+            .SetCreatureTags(TagsDefinitions.CreatureTagWildShape)
+            .SetSizeDefinition(Main.Settings.MakeLargeWildshapeFormsMedium
+                ? CharacterSizeDefinitions.Medium
```
ThyWoof, line 240: still not there. it's usually a bad idea to change blueprints like that. check how other large WS are handled when MakeLargeWildshapeFormsMedium changes. You need to add these names to a list accessed at SwitchMakeLargeWildshapeFormsMedium(). Also, aren't these blueprint names the same as the original ones? That won't work. Add a prefix as others.

ThyWoof:
after further reviewing I realized the mod changed them from Large to Medium if you enable the mentioned setting. Are they broken in vanilla game and set as Medium? They weren't.

ThyWoof:
I removed the original feature we had now that TA made all wildshape medium on purpose. The reason was to ensure no heroes would get locked out on smaller spaces. In fact they took UB cue and made it official. I don't believe we should mess with their sizes anymore.

PR #3459, opened 2023-06-09
Some tweaks

HIEROT:
1. Now eldritch smite can knock prone enemies Huge or smaller.
2. A mismatch in prerequisite of potent spellcaster is addressed.
3. Allow bard college of war dancer improved war dance to cast any cantrips with momentum and to kill enemies in all ways prolonging war dance. (This is put here cause I just implemented it for self-use and want to share it with others. Drop this feature if you wish)


ThyWoof:
We ended up removing potent spell caster on warlocks. Please only send the smite change on another request after merging latest dev.

PR #3462, opened 2023-06-09
Tweak War dance to allow non-melee attack regain momentum and non-mel…

HIEROT:
…ee kill prolong war dance
Just sharing this.


dolenam317:
The subclass is fine as it is. It doesn need any buf.

HIEROT:
> The subclass is fine as it is. It doesn need any buf.

fair enough, just pushed to my fork if anyone want it lol

ThyWoof:
Won't merge in respect to original author.

PR #3465, opened 2023-06-09
Some tweaks

EnderWiggin:


SolastaUnfinishedBusiness/CustomBuilders/InvocationsBuilders.cs
```diff
@@ -44,23 +44,25 @@ internal static InvocationDefinition BuildEldritchSmite()
...
         {
-            GameConsoleHelper.LogCharacterAffectedByCondition(defend_character,
-                        ConditionDefinitions.ConditionProne);
-            return new EffectForm[] {
-                    EffectFormBuilder.Create()
-                        .SetMotionForm(MotionForm.MotionType.FallProne)
-                        .Build() };
-        }
-        else
             return null;
```
sztrzask, line 56: Out of curiosity - why return null instead of empty enumerable? I don't know anything about Solasta game engine, but I'm usually a little bit worried whenever I see return null when an empty collection could work too?

EnderWiggin: empty enumerable still needs creation, null is basically a constant, so no new object is being created to be immediately discarded

sztrzask: Enumerable.Empty is cached I think? I've also never run into memory issues in Solasta. NRE's on the other hands are a plenty :D

Just my 2 cents, I'm not here to argue, I just wanted to point that out. Thanks for modding!

ThyWoof (approved):

PR #3471, opened 2023-06-10
WIP:Race oligath

otearaisu:
Added race Oligath
#### Level 1
* Attribute Increase - Strength +2
* Attribute Increase - Constitution +1
* Language - You know Common and Giant.
* Move - 6
* Powerful Build - You count your size as Large when determining your carrying capacity.
* Hot-Blooded - You have resistance to cold damage.
* Stone's Endurance - When taking damage, you can use your reaction to roll a d12 + Constitution modifier to reduce damage. You can use this power equal to your proficiency bonus, and regain all expended uses after a long rest.


ThyWoof (changes requested):
Are we sure we need 2 more interfaces? Can’t we use the existing ones that trigger before damage?

otearaisu:
> Are we sure we need 2 more interfaces? Can’t we use the existing ones that trigger before damage?

Need it to apply the reaction on damage receive so that the damage amount is already calculated before prompting the reaction.
Also the existing on damage receive only triggers when character is still alive.
The new also is for handling cases that damage kills the character.

ThyWoof:
@EnderWiggin , can you please review this one?

ThyWoof:
@EnderWiggin is working on a fix for the damage reduction. Wondering if we would be able to leverage that one setting damage reduction to 0.

SolastaUnfinishedBusiness/Properties/Resources.Designer.cs
```diff
@@ -1791,7 +1791,19 @@ public class Resources {
                 return ((byte[])(obj));
             }
         }
-        
+
```
EnderWiggin, line 1794: I see new resource for race icon is added, but can't find actual image

SolastaUnfinishedBusiness/Races/Oligath.cs
```diff
@@ -0,0 +1,260 @@
...
+            .SetUsesProficiencyBonus(ActivationTime.Reaction, RechargeRate.LongRest)
+            .SetReactionContext(ExtraReactionContext.Custom)
+            .AddToDB();
+
+        powerOligathStoneEndurance
+            .SetCustomSubFeatures(new PhysicalAttackBeforeHitConfirmedOnMeStoneEndurance(powerOligathStoneEndurance));
+        return powerOligathStoneEndurance;
+    }
+
+    private class PhysicalAttackBeforeHitConfirmedOnMeStoneEndurance : IDamageReceived, IBeforeDamageReceived
```
EnderWiggin, line 116: instead of this new custom feature it is better to extend and use `FeatureDefinitionReduceDamage`, preferably after PR #3473 is merged.

EnderWiggin:
>Need it to apply the reaction on damage receive so that the damage amount is already calculated before prompting the reaction (similar to kindred spirit share pain).

Why you need damage to already be calculated?

> Also the existing on damage receive only triggers when character is still alive.
The new also is for handling cases that damage kills the character.

For one, I'm not sure dead characters should react. And second - existing `FeatureDefinitionReduceDamage` triggers before damage is applied, so it can be used to reduce damage and prevent death.

EnderWiggin (changes requested):

SolastaUnfinishedBusiness/Races/Oligath.cs
```diff
@@ -0,0 +1,260 @@
...
+            {
+                yield break;
+            }
+
+            int totalReducedDamage = CalculateReducedDamage(rulesetDefender, damageAmount);
+            GameConsoleHelper.LogCharacterUsedPower(rulesetDefender, featureDefinitionPower);
+            rulesetDefender.UpdateUsageForPower(featureDefinitionPower, featureDefinitionPower.CostPerUse);
+
+            int prevHitpoints = StatusBeforeDamageMap[rulesetDefender.guid]?.HitPoints ?? 0;
+            int currentHitpoints = prevHitpoints + totalReducedDamage - damageAmount;
```
EnderWiggin, line 223: I have doubts this will work properly if character had any temporary HP before this damage instance

otearaisu:
> > Need it to apply the reaction on damage receive so that the damage amount is already calculated before prompting the reaction (similar to kindred spirit share pain).
> 
> Why you need damage to already be calculated?
This is similar to how the tabletop works. The damage is rolled before player decides to use reaction to reduce damage. Having it trigger before damage will make the feature less useful.
> 
> > Also the existing on damage receive only triggers when character is still alive.
> > The new also is for handling cases that damage kills the character.
> 
> For one, I'm not sure dead characters should react. And second - existing `FeatureDefinitionReduceDamage` triggers before damage is applied, so it can be used to reduce damage and prevent death.

Can try to rework, but see concern above. Also seems it currently not support rolled damage reduction?

Issue #3476, opened 2023-06-10
[1.5.52.1] Flanking does not work properly with creatures larger than medium


Andargor:
Currently, flanking does not provide advantage properly with creatures larger than medium, only certain squares seems to provide it. It should allow advantage according to the position of the attacker and at least one ally on the entire opposite sides of the creature, or diagonals.

I am still willing to review the code and propose a PR to implement it properly.

See this guide: https://www.dndlounge.com/flanking-5e/

Diagram showing where it applies or not (only medium and large shown, but should apply to Huge, Gargantuan and Colossal too):

![flanking-5e-squares-examples](https://github.com/SolastaMods/SolastaUnfinishedBusiness/assets/3580755/fe327dea-d0ad-4ea4-af86-0e31190dc101)

ThyWoof:
It won’t trigger on row 3, col 2 scenario. Which on above scenarios it isn’t working?

Andargor:
Just looking at the Large creatures, and skipping reach weapon tests:

### Should grant advantage, but doesn't

![mzReNz9](https://github.com/SolastaMods/SolastaUnfinishedBusiness/assets/3580755/0460b16f-4631-415a-be5d-036b9a8ce34f)

![KZ11C2L](https://github.com/SolastaMods/SolastaUnfinishedBusiness/assets/3580755/39cb1564-6ea3-4c82-bde1-aaccdec7ad54)
![WaIGDS8](https://github.com/SolastaMods/SolastaUnfinishedBusiness/assets/3580755/cdf6ebcf-05fc-4d82-a29e-d0198f7c4a18)

### Grants advantage but shouldn't

![WaIGDS8](https://github.com/SolastaMods/SolastaUnfinishedBusiness/assets/3580755/cdf6ebcf-05fc-4d82-a29e-d0198f7c4a18)

ThyWoof:
How tall is that monster? You’re flying. Depending on height attack is coming from a diagonal and won’t trigger.

ThyWoof:
Game is also a bit odd as it uses the height as well to determine sides. Can you test with races like human, elf, etc.?

Andargor:
Test #2, fairy flight dismissed, oddity in the diagonal. Monster is large, by definition it's a cube, so 10ft tall in this case.

### Works

[...]

PR #3494, opened 2023-06-11
Implementation of math-based flanking using attacker and ally centers

Andargor:
This implementation uses custom classes, since I recycled old code, they are stuffed as private classes into the flanking class. It could be refactored to use more native/Unity code.

I have tested impact on FPS, and it is minimal with your caching scheme. It seems to work with the reach weapons option.

EDIT: Ref: #3476


ThyWoof:
Thanks, it might take a while until I can review and merge this one as will be away from game computer until next weekend.

ThyWoof (approved):

PR #3510, opened 2023-06-13
Initial Commit to Eldritch Surge.

HIEROT:
Known issues:
Absent of Subclass Sigil and localized string automation and some battle logs.
TL;DR: save in the turn with Blast Reload causes minor problems.
Status of casted cantrip is unserializable though latent in memory and removed on battle turn end, so if the memory is lost (say you save between cast main cantrip, restart game and load) you lose cast bonus cantrip). Also if you sl game without ending your turn, the cached cantrips are still there and will prevent you from adding Blast Reload if it's not already there until your next turn end, and granting you cast bonus cantrip if you already have Blast Reload, though haven't yet cast main cantrips.


HIEROT:
Oh forget too add ranged spell attack no disadvantage within enemy melee range.

SolastaUnfinishedBusiness/Subclasses/PatronEldritchSurge.cs
```diff
@@ -0,0 +1,416 @@
...
+                0,
+                0
+                );
+        }
+
+
+    }
+    private sealed class HandleCastCantrip : IActionExecutionHandled, ICharacterTurnEndListener, IQualifySpellToRepertoireLine
+    {
+        private List<SpellDefinition> usedCantrips = new List<SpellDefinition>();
```
HIEROT, line 362: This private structure is unserielizable and causes minor problems.

HIEROT:
New Bing is truely my savior!
image

HIEROT:
@ThyWoof How to add png resource? I used default VS resx editor to import image, but I cannot edit its type to byte[]. Or Any automation script to run this?

ThyWoof:
> @ThyWoof How to add png resource? I used default VS resx editor to import image, but I cannot edit its type to byte[]. Or Any automation script to run this?

I manually edit resources.resx and then clean project.

ThyWoof:
https://github.com/SolastaMods/SolastaUnfinishedBusiness/blob/dev/SolastaUnfinishedBusiness/Properties/Resources.resx

HIEROT:
> https://github.com/SolastaMods/SolastaUnfinishedBusiness/blob/dev/SolastaUnfinishedBusiness/Properties/Resources.resx

Alright, having done the same... Now this subclass is completely implemented, excluding language localization other than EN and zh-CN.

SolastaUnfinishedBusiness/Api/GameExtensions/RulesetCharacterExtensions.cs
```diff
@@ -32,12 +32,16 @@ internal static bool IsWieldingTwoHandedWeapon([NotNull] this RulesetCharacter _
...
         var hero = character as RulesetCharacterHero ?? character?.OriginalFormCharacter as RulesetCharacterHero;
 
+        
         // required to ensure Tactician Adept feat doesn't increase dice for other fighter subclasses
         if (hero == null ||
             (hero.ClassesAndSubclasses.TryGetValue(klass, out var characterSubclassDefinition) &&
              characterSubclassDefinition.Name != subclass))
         {
-            return 1;
+            // Return 1 is very bad practice here! Making function name semantics different from actual effect!
```
EnderWiggin, line 41: While I agree that returning 1 here is bad practice, this change will break other features, unless there's according fix in other relevant places.

HIEROT: Of course I noticed the comment there but I just didn't figure out where this is needed,  so didn't take care of it_(:з」∠)_

HIEROT:
@EnderWiggin I thought there is a serielizable property to hold cast spells and the like to prevent stautus lost between sl and restart game. Do you have an idea? Thanks.

SolastaUnfinishedBusiness/Api/GameExtensions/RulesetCharacterExtensions.cs
ThyWoof: this bad practice was introduce to support gambits granted from feat to non Tactician.

ThyWoof (changes requested):

SolastaUnfinishedBusiness/CustomBuilders/InvocationsBuilders.cs
```diff
@@ -655,11 +655,12 @@ public IEnumerator OnActionFinished(CharacterAction action)
             }
 
             if (action.ActionType != ActionDefinitions.ActionType.Bonus &&
-                action.ActingCharacter.PerceptionState == ActionDefinitions.PerceptionState.OnGuard)
+                //action.ActingCharacter.PerceptionState == ActionDefinitions.PerceptionState.OnGuard
+                action.ActionDefinition.ActionScope == ActionDefinitions.ActionScope.Battle)
```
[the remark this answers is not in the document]
HIEROT: To better comply with other similar check for whether action is in combat. Besides, checking character on guard will result no extra buff applied when the party happens to be caught on fight, because at that time the party is no more on guard.

SolastaUnfinishedBusiness/Api/GameExtensions/RulesetCharacterExtensions.cs
HIEROT: I use this function to calculate the subclass level of Eldritch Surge to determine number of beams. If you are other subclass warlock,this will return 1 and causes you only having one beam. If it's inconvenient to modify here, it's fine to just add a private function handling this.

ThyWoof (approved):

PR #3514, opened 2023-06-13
Race Wendigo

otearaisu:
#### Level 1
* Strength +2
* Dexterity +1
* Stalker - You are proficient in Stealth
* Powerful Build - You count your size as Large when determining your carrying capacity.
* Natural Lunger - Your melee range increases by 5 feet. Does not stack with Lunger.
* Sucker Punch - On the first round of combat, if you hit a creature with an attack roll and it's initiative is lower than you, the creature takes additional 2d6 damage.


SolastaUnfinishedBusiness/Api/GameExtensions/EnumExtensions.cs
```diff
@@ -135,7 +135,8 @@ internal enum ExtraAdditionalDamageTriggerCondition
 
     TargetWithin10Ft = 9000,
     TargetIsDuelingWithYou = 9002,
-    FlurryOfBlows = 9003
+    FlurryOfBlows = 9003,
+    TargetHasLowerInitiativeOnFirstRound = 9004
```
ThyWoof, line 139: This is an old pattern in the mod that can be better accomplished with our interfaces. Please take a look at Roguish Slayer that has a very similar feature and try to follow that pattern. here: https://github.com/SolastaMods/SolastaUnfinishedBusiness/blob/f0f84219c08fbf961ad545fc647135a32e626e53/SolastaUnfinishedBusiness/Subclasses/RoguishSlayer.cs#L198

ThyWoof (changes requested):
Also that PNG looks ugly, lol. Sad we lost a very good graphics resource on the mod. Any changes you could try to use AI drawing software to achieve a better result. @HIEROT could guide you there.

otearaisu:
> Also that PNG looks ugly, lol. Sad we lost a very good graphics resource on the mod. Any chances you could try to use AI drawing software to achieve a better result. @HIEROT could guide you there.

Sadly, the AI drawing software is not available in my region. I'll try to find if there are other sources though.

SolastaUnfinishedBusiness/Api/GameExtensions/EnumExtensions.cs
otearaisu: Applied requested change.

HIEROT:
@otearaisu here you are.
image

HIEROT:
image
image

HIEROT:
Oh, I tried to improve them by adding more description.
image
image

ThyWoof:
Now we are talking

HIEROT:
I think this one is the best.
image

otearaisu:
> I think this one is the best. image

Amazing! Thanks for this haha originally planned to draw it manually

otearaisu:
Updated both png and implementation now.

ThyWoof (approved):

PR #3533, opened 2023-06-16
update chinese translation

magicskysword:


ThyWoof (changes requested):
I had dumped latest translations diagnostics already. Why they appear on this PR?

ThyWoof (approved):

PR #3571, opened 2023-06-20
Eldritch Surge tweaks and translations

HIEROT:


HIEROT:
Alright, I shall take care of this after all refactors. @ThyWoof

ThyWoof:
Sorry, I’m done now with changes. I’ll test to ensure nothing is broken and still works as designed.

ThyWoof:
Maybe remove that FeatureSet at level 18 and simply rephrase level 6 to state starting from 18th you always get an extra action and can stack up to 2.

HIEROT:
> Maybe remove that FeatureSet at level 18 and simply rephrase level 6 to state starting from 18th you always get an extra action and can stack up to 2.
> 
> gotta still add the additional action at 18. This way purist won’t complain about something appearing at 18th on progression list.

I wanted to do that but didn’t know how. It should be like what you said.

HIEROT:
Thank you for your hard work)

HIEROT:
In the beginning I check for condition so I need a lvl 18 feature. It now seems that it's fine just check lvl and remaining slots, no need for extra condition and feature at lvl 18. @ThyWoof

HIEROT:
After reviewing current code I realize that now lvl 18 will always grant you an action, but this is not intended. The logic is:
- 1. >= Lvl 18 && no remaining slot? Yes: 4 No: 2
- 2. Used activatable blast pursuit? Yes: 3
- 3. You have an extra action this turn, you can do anything listed in the feature with these actions.
- 4. You have 2 extra actions this turn, but you must cast EB to accquire them, one by one.
So to incorporate both at lvl 6 think we need a hidden permanent conditon/feature to handle action granting, and an activable power as a tag to mark whether you have activate lvl 6 feature.
I shall take care of this (if you haven't already @ThyWoof )

ThyWoof (approved):

PR #3591, opened 2023-06-22
Action switching

EnderWiggin:
Implemented ability to switch actions:
image

Controlled by `Settings.EnableActionSwitching` flag that is disabled by default and currently is not in the mod settings UI.


HIEROT:
Such a difficult and meaningful job.  You have my respect.

ThyWoof (approved):

PR #3611, opened 2023-06-23
Refactor on Eldritch Surge

HIEROT:
Received feedback from community of my language:
This subclass takes too long to warm up because of many powers and spells cost bonus action. Though we can apply them all before combat but it means sl for preparing an ambush, which is experience destoying.
levell 15~17 is too empty.
Also to clarify and simplify description, now I want to make these changes:
lvl 6 Blast Pursuit will grant you a main or bonus action (cannot attack with it), and initially lasts for 2 round, kill with EB prolongs it.
On lvl 16 and higher, this feature will grant you both, and even if you have no pact slots, you benefit from it.
Now I implement this as following


SolastaUnfinishedBusiness/Subclasses/PatronEldritchSurge.cs
```diff
@@ -50,6 +53,14 @@ internal class PatronEldritchSurge : AbstractSubclass
         .SetFeatures(AdditionalActionBlastPursuit)
         .AddToDB();
 
+    private static readonly ConditionDefinition ConditionExtraActionBlastPursuitOff = ConditionDefinitionBuilder
```
HIEROT, line 56: Add additional bonus action condition and feature

SolastaUnfinishedBusiness/Subclasses/PatronEldritchSurge.cs
```diff
@@ -320,48 +328,64 @@ private sealed class ExtraActionBlastPursuit : IActionExecutionHandled
...
+                    {
+                        if (warlockLevel < 16)
+                        {
+                            rulesetCharacter.RemoveAllConditionsOfType(ConditionExtraActionBlastPursuit.Name);
+                        }
+                        return;
+                    }
+                    break;
+                default:
+                    return;
```
HIEROT, line 370: Grant action of the same category after an action is finished, and remove the other action if you the action you finished is granted by this feature when lower than lvl 16

HIEROT: but the implementation here is far from perfect, because if you have actions from other sources, merely checking the action type will result in removing the other even if you haven't used the action granted by this feature
I know there is a way to check the source feature of your action but don't know how

ThyWoof: @EnderWiggin , wanna help here?

SolastaUnfinishedBusiness/Subclasses/PatronEldritchSurge.cs
```diff
@@ -375,7 +399,7 @@ private static void InflictCondition(RulesetCharacter rulesetCharacter)
                 0
             );
 
-            var title = ConditionExtraActionBlastPursuit.GuiPresentation.Title;
+            var title = ConditionBlastPursuit.GuiPresentation.Title;
```
HIEROT, line 402: ConditionExtraActionBlastPursuit has no localization

SolastaUnfinishedBusiness/Subclasses/PatronEldritchSurge.cs
HIEROT: I think I have figured it out. We can get the index of GameLocationCharacter.CurrentActionRankBytype which indicates the action used number. And then we get information from ActionSwitching.

HIEROT:
Now Blast Pursuit will always grants you an ma and ba, if your warlock level is less than 16, after you consume one,  the other will be removed.
And now the range is decreased based on level and condition instead of actions you have used. This is in accordance with the additional bonus action to reposition yourself with teleportation spells or invocations.
Now you elegantly blink here and there across the battlefield, bestow destruction and then safely retreat. Cool as hell)

ThyWoof (approved):

PR #3677, opened 2023-07-01
Fix Ability of the Chain Master Pseudo Dragon inflicting unsavable pa…

HIEROT:
…ralyze.

Modify it from fixed dc to spell dc

Note: After careful inspection the remove of
```
formsParams.activeEffect == null &&
```
has no other implications.
This is indeed a logic bug in vanilla game. When creating EffectForm there are detailed modifications to them, but when applying ConditionForm this method simply override saving info from the formsParam, which is not correct because this ignores subfeature overrides

### DIFF
--- a/SolastaUnfinishedBusiness/CustomBuilders/InvocationsBuilders.cs
+++ b/SolastaUnfinishedBusiness/CustomBuilders/InvocationsBuilders.cs
@@ -523,7 +523,13 @@ internal static class InvocationsBuilders
         var conditionAbilityPseudo = ConditionDefinitionBuilder
             .Create(ConditionDefinitions.ConditionFlying12, "ConditionAbilityPseudo")
             .SetGuiPresentation(Category.Condition, ConditionDefinitions.ConditionPactChainPseudodragon)
-            .AddFeatures(FeatureDefinitionAdditionalDamages.AdditionalDamagePoison_GhoulsCaress)
+            .AddFeatures(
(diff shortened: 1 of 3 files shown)


HIEROT:
The cause of this issue is that it override save from definition of Eldritch Blast, where the default save attribute is Dex and a paralyzed creature auto fails dex save.
In addition, there should be a Feedback/&AdditionalDamageLine locolization string, but I don't know where to put it fit.

HIEROT:
> The cause of this issue is that it override save from the definition of your attack, where if the default save attribute is Dex, a paralyzed creature auto fails dex save. In addition, there should be a Feedback/&AdditionalDamageLine locolization string, but I don't know where to put it fit.

This is fixed

ThyWoof:
> > The cause of this issue is that it override save from the definition of your attack, where if the default save attribute is Dex, a paralyzed creature auto fails dex save. In addition, there should be a Feedback/&AdditionalDamageLine locolization string, but I don't know where to put it fit.
> 
> This is fixed

so do we need or not a new localization string?

ThyWoof (approved):

HIEROT:
> > > The cause of this issue is that it override save from the definition of your attack, where if the default save attribute is Dex, a paralyzed creature auto fails dex save. In addition, there should be a Feedback/&AdditionalDamageLine locolization string, but I don't know where to put it fit.
> > 
> > 
> > This is fixed
> 
> so do we need or not a new localization string?

Now no need of new strings. I just reused poison.

PR #3693, opened 2023-07-03
Fix some missing particle effects

HIEROT:
including cool spirit shroud and two features in Eldritch Surge class

### DIFF
--- a/SolastaUnfinishedBusiness/Spells/SpellBuildersLevel03.cs
+++ b/SolastaUnfinishedBusiness/Spells/SpellBuildersLevel03.cs
@@ -462,9 +462,10 @@ internal static partial class SpellBuilders
                             .Create()
                             .SetConditionForm(ConditionDefinitionBuilder
                                 .Create($"Condition{SpiritShroudName}{damage}")
-                                .SetGuiPresentationNoContent(true)
+                                .SetGuiPresentationNoContent(false)
                                 .SetSilent(Silent.WhenAddedOrRemoved)
                                 .CopyParticleReferences(ConditionSpiritGuardiansSelf)
+                                .SetPossessive()
                                 .SetFeatures(FeatureDefinitionAdditionalDamageBuilder
                                     .Create($"AdditionalDamage{SpiritShroudName}{damage}")
                                     .SetGuiPresentationNoContent(true)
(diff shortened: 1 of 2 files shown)


SolastaUnfinishedBusiness/Spells/SpellBuildersLevel03.cs
```diff
@@ -462,9 +462,10 @@ internal static SpellDefinition BuildSpiritShroud()
                             .Create()
                             .SetConditionForm(ConditionDefinitionBuilder
                                 .Create($"Condition{SpiritShroudName}{damage}")
-                                .SetGuiPresentationNoContent(true)
+                                .SetGuiPresentationNoContent(false)
```
ThyWoof, line 465: This won’t make any change on sprites FYI. What is your goal here? For a sprite you need a proper text on condition as well as a sprite reference on SetGuiPresentation. Did you test this change?

ThyWoof (changes requested):

SolastaUnfinishedBusiness/Spells/SpellBuildersLevel03.cs
HIEROT: if you set it to false, this condition is completely hidden, including its spirit circling sprites

HIEROT: But this fix has its limitation though, now it will show no content or something (cause i am not using english) at tooltip.
I think the missing of sprites is due to vanilla game code modification. If no content and hidden is set, the particle effect will also be hidden.

HIEROT: but this change is tested.

HIEROT: And oh I got the two concepts mixed up. I mean particle effects.

ThyWoof: undo this change on SetGuiPresentation so I can merge pls.

ThyWoof (changes requested):

SolastaUnfinishedBusiness/Spells/SpellBuildersLevel03.cs
HIEROT: OK

ThyWoof (approved):

PR #3698, opened 2023-07-03
Avoid broken UI when user disables filtering/sorting

Kiloku:
Properly remove/reapply all UI changes when disabling/enabling Inventory Filtering and Sorting


Kiloku:
I could swear I had written a proper commit message. Anyway, from my testing this is working as intended now.
Toggling the setting adjusts the UI at runtime. There's just one small wonkiness that remains (the bottom section might get slightly cut-off or overlaid, depending on which state you're toggling from/to) if the inventory is open at the exact moment you toggle, but just closing and opening the inventory fixes it.

ThyWoof (approved):

PR #3703, opened 2023-07-04
Serializable Custom Conditions

HIEROT:
Currently many homebrew features require storing some dynamic variables to compute for different purposes. And to make them persistent in game save, we just like stack conditions and count number, which is just not so versatile.
With patching in serialization process in RulesetActor, it now supports serialize objects inheriting RulesetCondition. This is an existing practice elsewhere in vanilla game.
So now if you have some condition variables (int, string, whatever) to store, you just create a dummy ConditionDefinition for reference and a new class inheriting RulesetCondition and ISerializable, implementing two serialization functions to to serialize your newly added properties, that's done.
we can copy/paste InflictCondition to make it usages similar to old practice, like append another param to initialize custom condition properties.
Have no impact on current saves. 
Hoping this utility helps with bolder imagination.

### DIFF
--- a/SolastaUnfinishedBusiness/Subclasses/PatronEldritchSurge.cs
+++ b/SolastaUnfinishedBusiness/Subclasses/PatronEldritchSurge.cs
@@ -531,14 +531,7 @@ internal class PatronEldritchSurge : AbstractSubclass
                 BaseDefinition.SerializeDatabaseReferenceList<SpellDefinition>(serializer, "CantripsUsedThisTurn", "SpellDefinition", cantripsUsedThisTurn);
                 if (serializer.Mode == Serializer.SerializationMode.Read)
                 {
-                    for (int i = 0; i < this.cantripsUsedThisTurn.Count; i++)
-                    {
-                        if (this.cantripsUsedThisTurn[i] == null)
-                        {
-                            this.cantripsUsedThisTurn.RemoveAt(i);
-                            i--;
-                        }
-                    }
+                    cantripsUsedThisTurn.RemoveAll(x => x is null);
                 }
             }
             catch (Exception ex)
--- a/SolastaUnfinishedBusiness/Subclasses/PatronEldritchSurge.cs
+++ b/SolastaUnfinishedBusiness/Subclasses/PatronEldritchSurge.cs
@@ -94,6 +94,11 @@ internal class PatronEldritchSurge : AbstractSubclass
 
     // LEVEL 14 Blast Reload
 
+    private static readonly ConditionDefinition ConditionBlastReloadSupport = ConditionDefinitionBuilder
+        .Create($"Condition{Name}BlastReloadSupport")
+        .SetGuiPresentationNoContent(true)
+        .AddToDB();
+
     public static readonly FeatureDefinition FeatureBlastReload = FeatureDefinitionBuilder
         .Create($"Feature{Name}BlastReload")
         .SetGuiPresentation(Category.Feature)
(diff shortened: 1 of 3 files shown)


HIEROT:
Also, a new TypeExtensions is added to handle patching generic overloaded functions, which makes matching functions easier.

EnderWiggin:
How serialization would handle changes to the custom RulesetConsition?

EnderWiggin:
Also since RulesetConditions seem to serialize their attributes in theory we can just add custom attributes to regular ruleset conditions to track some stuff

HIEROT:
> How serialization would handle changes to the custom RulesetConsition?

To my knowledge if the condition is not currently applied to any character, it will be fine.
If you newly add a property and properly handle it in default constructor, It will be fine too.
But a remove of property will cause nre when loading from save, but it is catched, so the game will not crash at the moment. But I don't know if it prevents new saving.
(All theoretical analyzation)

HIEROT:
> Also since RulesetConditions seem to serialize their attributes in theory we can just add custom attributes to regular ruleset conditions to track some stuff

no the serialization is hard coded, you need to hack its function to serialize other things. So I think it's better to make your own.

HIEROT:
I think we can add a RulesetCondition to a new category managed by mod so it supports some permanent custom variables)

HIEROT:
> How serialization would handle changes to the custom RulesetConsition?

But when unpdating your custom condition, you can do through multiple releases. Just postpone serialization by one release.

SolastaUnfinishedBusiness/Subclasses/PatronEldritchSurge.cs
```diff
@@ -480,13 +503,90 @@ public void OnCharacterTurnStarted(GameLocationCharacter locationCharacter)
...
+
+        public override void SerializeElements(IElementsSerializer serializer, IVersionProvider versionProvider)
+        {
+            base.SerializeElements(serializer, versionProvider);
+            try
+            {
+                BaseDefinition.SerializeDatabaseReferenceList<SpellDefinition>(serializer, "CantripsUsedThisTurn", "SpellDefinition", cantripsUsedThisTurn);
+                if (serializer.Mode == Serializer.SerializationMode.Read)
+                {
+                    for (int i = 0; i < this.cantripsUsedThisTurn.Count; i++)
```
EnderWiggin, line 534: while this loop seems to be correct, it is better to use RemoveAll() instead of manually looping

EnderWiggin:
[SerializationVersion(uint)] attribute can be used to specify version that can be retrieved and checked though `versionProvider` that's passed to serialization methods, but better to check if it actually works properly - meaning affects only each custom class and doesn't modify base class version. Most likely it works fine, but better safe than sorry.

HIEROT:
> [SerializationVersion(uint)] attribute can be used to specify version that can be retrieved and checked though `versionProvider` that's passed to serialization methods, but better to check if it actually works properly - meaning affects only each custom class and doesn't modify base class version. Most likely it works fine, but better safe than sorry.

as for this one, base serialization method is called beforehand. I agree we can use SerializationVersion to control but it's related to vanilla game version so we don't need to check it for our inheritances. Maybe modify this interface to set mod version.

EnderWiggin:
> > [SerializationVersion(uint)] attribute can be used to specify version that can be retrieved and checked though `versionProvider` that's passed to serialization methods, but better to check if it actually works properly - meaning affects only each custom class and doesn't modify base class version. Most likely it works fine, but better safe than sorry.
> 
> as for this one, base serialization method is called beforehand. I agree we can use SerializationVersion to control but it's related to vanilla game version so we don't need to check it for our inheritances. Maybe modify this interface to add mod version.

After quick look at the TA's code it seems that adding this attribute to our custom class would work properly - each version attribute is processed per _type_, so us adding and changing it shouldn't affect base code, except base serialization of the RulesetCondition maybe. Need to do a test to check if it is so.

HIEROT:
> > > [SerializationVersion(uint)] attribute can be used to specify version that can be retrieved and checked though `versionProvider` that's passed to serialization methods, but better to check if it actually works properly - meaning affects only each custom class and doesn't modify base class version. Most likely it works fine, but better safe than sorry.
> > 
> > 
> > as for this one, base serialization method is called beforehand. I agree we can use SerializationVersion to control but it's related to vanilla game version so we don't need to check it for our inheritances. Maybe modify this interface to add mod version.
> 
> After quick look at the TA's code it seems that adding this attribute to our custom class would work properly - each version attribute is processed per _type_, so us adding and changing it shouldn't affect base code, except base serialization of the RulesetCondition maybe. Need to do a test to check if it is so.

after these tweaks, my game save works just fine.

ThyWoof (approved):

PR #3757, opened 2023-07-09
Implement suspension of temporary flight from items and spells

Andargor:
From discussion #3709
[Andargor](https://github.com/Andargor) started this conversation in [Ideas](https://github.com/SolastaMods/SolastaUnfinishedBusiness/discussions/categories/ideas)

Here is my proposal, hopefully I didn't make too much of a mess of things. Maybe some things are not in the right place, but I was running in some sync issues (e.g. condition wasn't created in time for building the action)

- Shows an icon to land when a temporary flight condition is present, such as Winged Boots or the Fly spell
- The associated power stores the flight condition and the remaining duration
- Adds a flight suspended condition with the remaining time
- Shows a take off icon if flight is suspended and a condition display
- Taking off reinstates the flight condition and remaining duration is taken into account
- Does not work for wildshapes
- Does not affect races with permanent flying ability
- Only English localization done
- Setting in Gameplay > Rules: "Allow temporarily suspending flight..."


EnderWiggin:
how does it work with concentration? If I stop flight, will concentration on it drop? will it resume? Or after resuming I can fly without worrying of being hit and falling?

ThyWoof:
> > how does it work with concentration? If I stop flight, will concentration on it drop? will it resume? Or after resuming I can fly without worrying of being hit and falling?
> 
> That's a good point, in the case of Fly the concentration is removed on suspend, I'll look at how to restore it as well. The concentration should remain for both the suspended and restored state, to be clear.

Have you looked into this one?

Andargor:
> > > how does it work with concentration? If I stop flight, will concentration on it drop? will it resume? Or after resuming I can fly without worrying of being hit and falling?
> > 
> > 
> > That's a good point, in the case of Fly the concentration is removed on suspend, I'll look at how to restore it as well. The concentration should remain for both the suspended and restored state, to be clear.
> 
> Have you looked into this one?

I have a solution in mind, by tracking the suspended condition it keeps concentration. I just haven't implemented and tested it yet, since it can get complicated in the case that the Fly spell was cast by someone else.

I am limited on time in the next few days, I might finish this weekend.

Andargor:
I've implemented concentration tracking for suspended flight. Cases tested:

- Item effect, suspend and resume
- Spell effect with concentration, suspend, resume
- Break/cancel concentration while suspended and active
- Metamagic (e.g. Twinned Fly) with various combinations of casters and targets, concentration broken while suspended, and when one target flying and another not
- Save/load while suspended

I've implemented try/catch blocks in case an edge case breaks something.

SolastaUnfinishedBusiness/Api/GameExtensions/RulesetCharacterExtensions.cs
```diff
@@ -560,4 +560,20 @@ internal static RulesetCharacterHero GetOriginalHero(this RulesetCharacter chara
     {
         return character as RulesetCharacterHero ?? character.OriginalFormCharacter as RulesetCharacterHero;
     }
+
+    internal static bool HasTemporaryConditionOfType(this RulesetCharacter character, string conditionName)
+    {
+        foreach (KeyValuePair<string, List<RulesetCondition>> item in character.ConditionsByCategory)
```
ThyWoof, line 566: why not traverse character.AllConditions to avoid the double loop?

SolastaUnfinishedBusiness/Displays/RulesDisplay.cs
```diff
@@ -61,6 +61,12 @@ internal static void DisplayRules()
             Main.Settings.DontEndTurnAfterReady = toggle;
         }
 
+        toggle = Main.Settings.AllowFlightSuspend;
+        if (UI.Toggle(Gui.Localize("ModUi/&AllowFlightSuspend"), ref toggle, UI.AutoWidth()))
```
ThyWoof, line 65: not sure if this should be under Rules or Game UI.

SolastaUnfinishedBusiness/Translations/en/Settings-en.txt
```diff
@@ -20,6 +20,7 @@ ModUi/&AllowBeardlessDwarves=Allow beardless <color=#D89555>Dwarves</color>
 ModUi/&AllowClubsToBeThrown=Allow <color=#D89555>clubs</color> to be thrown like <color=#D89555>daggers</color> or <color=#D89555>hand axes</color>
 ModUi/&AllowDruidToWearMetalArmor=Allow <color=#D89555>Druid</color> to wear metal armor
 ModUi/&AllowDungeonsMaxLevel20=Allow dungeons with max level 20
+ModUi/&AllowFlightSuspend=Allow temporarily suspending flight from items and spells (does not affect wildshape or flying races)
```
ThyWoof, line 23: follow UI pattern here and instead of parentheses use `<i><color=#F0DAA0>[` and `]</color></i>` instead

ThyWoof (changes requested):
please check comments.

Andargor:
Noted, will correct tonight

Andargor:
Changed done, good to go

ThyWoof (approved):

PR #3848, opened 2023-07-25
race-oligath - fixed stone endurance behaviour

otearaisu:
- check available reactions activating power
- show remaining usable power on react prompt
- refactored to use FeatureDefinitionReduceDamage

### DIFF
--- a/SolastaUnfinishedBusiness/Api/GameExtensions/GameLocationCharacterExtensions.cs
+++ b/SolastaUnfinishedBusiness/Api/GameExtensions/GameLocationCharacterExtensions.cs
@@ -168,13 +168,8 @@ public static class GameLocationCharacterExtensions
                && !character.HasConditionOfType(RuleDefinitions.ConditionParalyzed);
     }
 
-    internal static bool CanReact(this GameLocationCharacter instance, bool ignoreReactionUses = false)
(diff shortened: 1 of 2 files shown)


SolastaUnfinishedBusiness/Api/GameExtensions/GameLocationCharacterExtensions.cs
```diff
@@ -168,13 +168,8 @@ internal static bool CanAct(this GameLocationCharacter instance)
                && !character.HasConditionOfType(RuleDefinitions.ConditionParalyzed);
     }
 
-    internal static bool CanReact(this GameLocationCharacter instance, bool ignoreReactionUses = false)
+    internal static bool IsReactionAvailable(this GameLocationCharacter instance, bool ignoreReactionUses = false)
```
ThyWoof, line 171: Why do you need this change? Game code does the same. Characters can only react if they can act. What is the special scenario?

ThyWoof (changes requested):

SolastaUnfinishedBusiness/Api/GameExtensions/GameLocationCharacterExtensions.cs
otearaisu: In order to activate Stone Endurance when prone. I think this is different from other actions as prone would just disallow you from attacking.

ThyWoof (approved):

PR #4330, opened 2024-02-04
update russian translation

Kranay:


ThyWoof (changes requested):
This is using an older file version. Cannot merge

PR #4430, opened 2024-03-16
Chinese translation correction

zxcvbnm61004:


ThyWoof (changes requested):
Please only submit Chinese translations. You are changing EN ones here without any reason.

zxcvbnm61004:
> Please only submit Chinese translations. You are changing EN ones here without any reason.

OK.

zxcvbnm61004:
In Bolgrif-en.txt, Firbolg has not been changed to Bolgrif.
In Malakh-en.txt, Aasimar has not been changed to Malakh.
DomainSmith-en.txt does not make changes.

PR #4443, opened 2024-03-21
Make Ring of Regeneration closer to RAW

PhilPJL:
An outstanding bug noted by Artyoan is that the Ring of Regen is way overpowered and heals 2hp per round.  This changes it to 1hp per minute which is closer to 1d6 per 10 minute RAW.
However should this be optional or not?


PhilPJL:
Need to remove log statement.

PhilPJL:
It may be better to add a UB version of the ring with the correct stats

ThyWoof (approved):

PR #4461, opened 2024-03-26
Tweaks

PhilPJL:
Remove log statement.
Change description to use 'round' not 'turn'.


ThyWoof (commented):
Wouldn’t this fully override original desc? Might be better to follow same pattern I did on Barb Reckless buff debuff setting.

PhilPJL:
The original text was 

Equipment/&BeltOfRegeneration_Function_Description=Regenerate 5 hit points per **turn** for one minute.

which doesn't agree with 

Equipment/&DLC1_Item_BeltOfRegeneration_Description=Activate this belt to regenerate 5 HP per **round** for one minute, as long as you are above 1 HP, and are wearing and attuned to the belt. This power can be used once per day.

Turn isn't a measure of time, round is.

ThyWoof (approved):

PR #4462, opened 2024-03-27
Fix Versatility Adept and imporve description

HIEROT:


SolastaUnfinishedBusiness/Subclasses/Builders/EldritchVersatility.cs
```diff
@@ -80,11 +80,17 @@ public enum PointUsage
             .Setup(InvocationPoolTypeCustom.Pools.EldritchVersatilityPool, 1, true)
             .AddToDB();
 
+    public static readonly FeatureDefinition FeatureEldritchVersatilityGrantPoolAndSwitch = 
+        FeatureDefinitionBuilder
+            .Create($"Feature{Name}GrantPoolAndSwitch")
+            .SetGuiPresentation(Category.Feature, hidden: true)
```
ThyWoof, line 86: no need for a GUI Presentation here. Use SetGuiPresentationNoContent(true)

ThyWoof (changes requested):
check minor change request

HIEROT:
@ThyWoof done

ThyWoof (approved):

HIEROT:
please revert the last commit as it was a mistake

PR #4641, opened 2024-05-22
Update CureWounds.json

kdgmo:
Healing has been buffed in one D&D.
The buffed content increases the recovery dice from 1 to 2, and the dice when upcasting also increases from 1 to 2.


ThyWoof:
JSON file changes won’t make any difference.

kdgmo:
So which file should I edit?

PR #4646, opened 2024-05-22
Update Regenerate.json

kdgmo:
This order was a mistake by the Solastar production team.  It is a magic that restores amputated body parts.
However, since there is no effect of amputated body parts in Solasta, this spell has become a spell with a lower healing amount than the previous level magic, Recovery. This is clearly a mistake on the part of the Solastar production team.
So I think a buff is needed.


ThyWoof:
JSON files are only static dumps I use for reference. They aren’t used at all in mod. Please open an issue stating the required change and I’ll research.

kdgmo:
Which file should I edit? The problem with this spell is that the recovery amount is obviously too low compared to lower level spells. In the first place, it was a spell used to restore amputated body parts, not a spell written based on the amount of recovery.

PR #4651, opened 2024-05-24
ONE D&D Heal buff [CureWounds]

kdgmo:
Healing has been buffed in one D&D.
The buffed content increases the recovery dice from 1 to 2, and the dice when upcasting also increases from 1 to 2.
The content of ONE D&D has been implemented as is.
But this is my first time working on it. Is it right to work like this?


ThyWoof:
this code doesn't even compile.

kdgmo:
this code doesn't even compile.
I'm a beginner, and compilation isn't working right now. I don't know what the problem is.  Severity Code Description Project File Line Suppression Status
Warning MSB3245 This reference cannot be resolved. Assembly "Assembly-CSharp" could not be found. Make sure the assembly is on disk. If your code requires this reference, a compilation error may occur.    SolastaUnfinishedBusiness C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets 2084
What should I do to fix this error?

PR #4655, opened 2024-05-28
ONE D&D Heal buff [CureWounds]

kdgmo:
ONE D&D Heal buff [CureWounds]


EnderWiggin:
It looks like you forgot to push most of your changes, or am I missing something? I can only see 1 line changed, and that line just declares new property.

kdgmo:
> ush most of your

I didn't know how to upload them together, so I uploaded them separately.SolastaUnfinishedBusiness/Models/SrdAndHouseRulesContext.cs
This is the function where you can adjust the numerical values ​​in earnest.

PR #4659, opened 2024-05-29
HealSpell buff

kdgmo:
Healing spells have been strengthened to match ONE D&D.
CureWounds buffed content increases the recovery dice from 1 to 2, and the dice when upcasting also increases from 1 to 2.
HealingWord buffed content increases the recovery dice from 1 to 2, and the dice when upcasting also increases from 1 to 2.
PrayerOfHealing buffed content increases the recovery dice from 2 to 3, and the dice when upcasting also increases from 1 to 2.
MassHealingWord buffed content increases the recovery dice from 1 to 3, and the dice when upcasting also increases from 1 to 2.
MassCureWounds buffed content increases the recovery dice from 3 to 5, and the dice when upcasting also increases from 1 to 2.


ThyWoof:
incomplete. added on my end.

PR #4690, opened 2024-06-07
queue attack created through attack after magic effect to fix #4679

### LINKED ISSUE #4679: Several quicken spell issues
1. If a sorcerer character is hasted, after doing the first action to cast a cantrip, I cannot use sorcery points to bonus action quicken a spell. 
I'm just not able to quicken a spell in general after my action,

2. After quickening a cantrip before using my action, I cannot then use my main action to cast a leveled spell (which I think I should be able to according to the rules).

3. After quickening a cantrip (only tried booming blade), I can then repeat the quicken metamagic for another cantrip infinitely (well, until I ran out of sorcery points at least.) So I got 3 bonus actions in one round.

EnderWiggin:
without queueing, current action (casting in this case) is interrupted without calling upkeep method that spends action

### DIFF
--- a/SolastaUnfinishedBusiness/Patches/CharacterActionMagicEffectPatcher.cs
+++ b/SolastaUnfinishedBusiness/Patches/CharacterActionMagicEffectPatcher.cs
@@ -539,7 +539,7 @@ public static class CharacterActionMagicEffectPatcher
                     else
                     {
                         ServiceRepository.GetService<IGameLocationActionService>()?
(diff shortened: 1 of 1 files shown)


EnderWiggin:
This solution seems to have some problems - need more time to test and fix.

ThyWoof:
fixed https://github.com/SolastaMods/SolastaUnfinishedBusiness/commit/435393e9712d35439f312fe6ba06d76de6ff74f0

PR #4701, opened 2024-06-09
fix a bug with campaign translate

magicskysword:
When I use campaign translator to translate [The West Marches of Solasta (Level 1-20)](https://steamcommunity.com/sharedfiles/filedetails/?id=3173824542) , I found the progress stuck at 94% and find a exception in log.

```
NullReferenceException: Object reference not set to an instance of an object
  at UserItem.get_DocumentFragments () [0x00033] in <14069e0776c443f6af77ee3eeee86525>:0 
  at SolastaUnfinishedBusiness.Models.TranslatorContext+TranslatorBehaviour+<TranslateUserCampaignRoutine>d__7.MoveNext () [0x008a5] in C:\Users\ThyWolf\Repos\SolastaUnfinishedBusiness\SolastaUnfinishedBusiness\Models\TranslatorContext.cs:668 
  at UnityEngine.SetupCoroutine.InvokeMoveNext (System.Collections.IEnumerator enumerator, System.IntPtr returnValueAddress) [0x00026] in <85d1d3e7744a4a47b5f51883bf40bba2>:0 
```

After breakpoint debugging, I found this Bug is due to get `UserItem.DocumentFragments`, needs to read `referenceItemDefinition.DocumentDescription`. But some reference items don't have a DocumentDescription (according to my breakpoint, the item is `CELongMace+3`).

My temporary solution is to skip the items, not sure what the effect will be, but let the translation finish for now.


ThyWoof:
Wouldn’t be better to add the missing data to these items? They are all UB

magicskysword:
I tried to add the default `documentDescription` setting for `ItemDefinitionBuilder`. And I've found that `userCampaignMapNode.userLocationName` is must be translated, because this name is actually site ID, If the map node's location name is different with site, the game will be wrong.

ThyWoof (approved):

PR #4789, opened 2024-07-08
Fix Psi Warrior's Force of Will to actually use INT

Swofty:
> Psi Warrior
>
> * Force of Will
>
> Your psionic energy grants you extraordinary resilience. At the start of each of your turns, you gain temporary hit points equal to your Intelligence modifier (minimum of 1) if you have at least 1 hit point. In addition, **you can use your Intelligence modifier instead of your Wisdom and Charisma modifier for saving throws if it's higher.**

Currently, when Force of Will is triggered, all that happens is that the displayed saving throw is shown as INT, but it didn't actually use the INT mod. This change fixes that.

image

However, several caveats:

(1) This change considers your saving throw proficiency when doing the replacement. It compares INT [+Prof] against the original WIS [+Prof] or CHA [+Prof], rather than just INT vs WIS/CHA. I like this more because if you get proficiency through sub-classing or feats, you will gain benefit or not lose out just because the base INT vs WIS works differently.

If we'd prefer just INT vs WIS, I have a code path in there to do just that. (Set `useNewBonuses = false`)

(2) The base game Mana Painter Sorcerer has Mana Absorption, which has similar wording:

> Your Charisma modifier is used (if higher) instead of other ability modifiers on your saving throws against spells and other magic effects.

This vanilla feature works by replacing the saving throw entirely to be Charisma if the CHA mod is greater than the original saving throw's ability mod. (This is done through an `ISavingThrowAffinityProvider` in `RulesetImplementationManager.TryRollSavingThrow`).This has interesting consequences, such as the fact that you get CHA proficiency added (to all saving throws basically), you don't benefit from cover against spells that were originally DEX saves, and that bonuses to CHA saves (like gnome adv on INT/WIS/CHA) apply for things like fireball.

This change does not do that. Rather, the save just swaps the ability scores, but is mostly still treated as a saving throw of the original type.

(3) Despite point 2, the saving throw type still does get changed to INT after Force of Will is applied, so any mod power that uses `OnSavingThrowInitiated` will race against it. This is currently not an issue because no Unfinished Business feat applies a bonus to INT/WIS/CHA through that handler. However, if this code will be reused for Barbarian Path of Savagery, then the race needs to be addressed for the Shield Master feat.

>Path of the Savagery
>
> * Furious Defense
>
> Starting at 6th level, while you are raging, **whenever you roll a Dexterity saving throw, you can use your Strength saving throw instead**. You gain +2 AC while you are raging and dual-wielding two melee weapons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Do you have a mod you want to see included here? We are happy to take new contri
 # How to compile
 
 0. Install all required development pre-requisites:
-    - [Visual Studio 2022 Community Edition](https://visualstudio.microsoft.com/downloads/)
+    - [Visual Studio 2022 Community Edition 17.8+](https://visualstudio.microsoft.com/downloads/)
     - [.NET x64 6.00.300 SDK](https://dotnet.microsoft.com/download/visual-studio-sdks)
 1. Download and install [Unity Mod Manager (UMM)](https://www.nexusmods.com/site/mods/21)
 2. Execute UMM, Select Solasta, and Install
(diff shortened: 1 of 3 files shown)


ThyWoof:
Thanks for the suggestions but will probably take the vanilla approach and remove older modifier from list, add new one. Meanwhile, I’m the solo developer around. If you could give a hand on why the hell remoraz aren’t dealing the reactive damage. Thanks

PR #4793, opened 2024-07-09
Tweak Reverse Gravity to not grant flying to creatures.

HIEROT:
This spell should not grant flying ability according to [Reverse Gravity](https://www.dndbeyond.com/spells/2233-reverse-gravity?msockid=242cc9d5fa03644c3642daabfbd165cc), otherwise it's useless as a 7th level CC spell.


ThyWoof:
In that case you should also change the parent condition to levitate. Please push this change.

HIEROT:
Flying is more propriate as Levitate will fix speed to 0, which will make a creature that is already flying not able to move at all.

ThyWoof (approved):

PR #4833, opened 2024-07-28
update russian translation

Kranay:


ThyWoof (approved):

ThyWoof:
Did a final change on sorcerer translations after QA review. Should release today in case you can update them

PR #4890, opened 2024-08-19
Major refactor on Patron Eldritch Surge

HIEROT:
Revised this subclass to make it less monotonous than EB the hell out)
[RESPEC REQUIRED]

changes:

Lvl 6 Blast Pursuit now automatically grant Eldritch Points at turn start so that you can use your turn to do other things with less loss.

Lvl 10 Blast Reload no longer allows 2 EB per turn, but you can cast a spell and EB with bonus action, or EB and cast a main action spell.

Battlefield Shorthand now allows you to take another shorthand when you cast the copied spell. so if you are lucky enough or with sufficient skill you can use them multiple times.

Battlefield Conversion's DC increased but fixed.

Eldritch Ward now grant AC bonus to a target for a round, just as Eldritch Aegis.


HIEROT:
oh pls clean up code on your side as I am not sure the settings @ThyWoof

PR #5025, opened 2024-11-01
Change Improved Abjuration to modify existing spell EffectForms

BShennette:
Initial version was copying power definition from the Hive Patron -- I didn't realize that was a one-time-use w/ no slot.
Now it modifies Counterspell and Dispel Magic spells for the caster. I tested it, and Counterspell uses the proficiency bonus for the Abjurer now, and continues to not for any character without this feature.

Vanilla log for dispel magic doesn't show an ability check in the log, so it's hard to judge, but it still works to dispel effects from a creature and the first form should be correct as it's the same as Counterspell's.

Also, this change alerted me to a bug that I think is from Vanilla? You can select proxy effects to dispel (like _Wall of Fire_ or _Flaming Sphere_) and an animation plays, but the effect doesn't end up getting dispelled. Might be worth looking into (not that I'm aware of any enemies that use that kind of spell atm?)

### DIFF
--- a/SolastaUnfinishedBusiness/Subclasses/WizardAbjuration.cs
+++ b/SolastaUnfinishedBusiness/Subclasses/WizardAbjuration.cs
@@ -122,49 +122,24 @@ public sealed class WizardAbjuration : AbstractSubclass
 
         ////////
         // Lv.10 Improved Abjuration
-        var powerCounterSpell = FeatureDefinitionPowerBuilder
+        var powerCounterSpellAffinity = FeatureDefinitionPowerBuilder
             .Create($"Power{Name}CounterSpell")
             .SetGuiPresentationNoContent(true)
-            .AddCustomSubFeatures(ModifyPowerVisibility.Hidden)
-            .SetUsesFixed(ActivationTime.Reaction, RechargeRate.ShortRest)
-            .SetEffectDescription(EffectDescriptionBuilder.Create()
-                .SetDurationData(DurationType.Instantaneous, 1)
-                .SetTargetingData(Side.Enemy, RangeType.Distance, 12, TargetType.IndividualsUnique)
-                .SetParticleEffectParameters(Counterspell)
-                .SetEffectForms(EffectFormBuilder.Create()
-                    .SetCounterForm(CounterForm.CounterType.InterruptSpellcasting, 3, 10, true, true)
-                    .Build())
-                .Build())
+            .AddCustomSubFeatures(new ModifyCounterspellAddProficiency())
             .AddToDB();
 
-        var powerCounterDispel = FeatureDefinitionPowerBuilder
+        var powerCounterDispelAffinity = FeatureDefinitionPowerBuilder
             .Create($"Power{Name}CounterDispel")
             .SetGuiPresentationNoContent(true)
-            .AddCustomSubFeatures(ModifyPowerVisibility.Hidden)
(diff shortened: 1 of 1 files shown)


HiddenHax:
Do Cloudkill, Insect Swarm and Black Tentacles count as proxies? If so, there are some casters that use these (Master Madracht and Mummy Lord iirc).

SolastaUnfinishedBusiness/Subclasses/WizardAbjuration.cs
```diff
@@ -627,6 +602,52 @@ void ReactionValidated()
         }
     }
 
+    private sealed class ModifyCounterspellAddProficiency() : IModifyEffectDescription
+    {
+        EffectDescription IModifyEffectDescription.GetEffectDescription(BaseDefinition definition, EffectDescription effectDescription, RulesetCharacter character, RulesetEffect rulesetEffect)
+        {
+            return EffectDescriptionBuilder
```
ThyWoof, line 609: no need to build a new effect description here. we already take care of having a copy that is passed to this method. Also a good idea to have the effect form as a static field to avoid building every time.

SolastaUnfinishedBusiness/Subclasses/WizardAbjuration.cs
```diff
@@ -627,6 +602,52 @@ void ReactionValidated()
...
+            return character.GetClassLevel(CharacterClassDefinitions.Wizard) >= 10 &&
+                definition == Counterspell;
+        }
+    }
+
+    private sealed class ModifyDispelMagicAddProficiency() : IModifyEffectDescription
+    {
+        EffectDescription IModifyEffectDescription.GetEffectDescription(BaseDefinition definition, EffectDescription effectDescription, RulesetCharacter character, RulesetEffect rulesetEffect)
+        {
+            return EffectDescriptionBuilder.Create(DispelMagic)
```
ThyWoof, line 628: no need to build a new effect description here. we already take care of having a copy that is passed to this method. Also a good idea to have the effect form as a static field to avoid building every time.

SolastaUnfinishedBusiness/Subclasses/WizardAbjuration.cs
```diff
@@ -627,6 +602,52 @@ void ReactionValidated()
...
+                .SetEffectForms(
+                    EffectFormBuilder.Create()
+                        .SetCounterForm(CounterForm.CounterType.DissipateSpells, 3, 10, true, true)
+                        .SetCreatedBy(true)
+                        .SetBonusMode(AddBonusMode.None)
```
ThyWoof, line 633: no need to set default values

SolastaUnfinishedBusiness/Subclasses/WizardAbjuration.cs
```diff
@@ -627,6 +602,52 @@ void ReactionValidated()
...
+                        .Build(),
+                    EffectFormBuilder.Create()
+                        .SetAlterationForm(AlterationForm.Type.DissipateSpell)
+                        .SetCreatedBy(true)
+                        .SetBonusMode(AddBonusMode.None)
```
ThyWoof, line 638: no need to set default values

ThyWoof (changes requested):

ThyWoof (approved):

PR #5067, opened 2024-11-20
Enable InvocationPactBlade to use CHA for weapon attacks

BShennette:
and Fix Hexblade to work with InvocationPactBlade for 2H weapons.

2024 Pact of the Blade invocation also allows you to use CHA with the bonded weapon. I've added the same qualifier from Hexblade to the invocation.

I also fixed Hexblade condition only looking for the Pact of the Blade _feature_, which doesn't exist if you enable the 2024 Pact invocations option; it will now also consider the invocation.

Untitled

### DIFF
--- a/SolastaUnfinishedBusiness/Subclasses/PatronSoulBlade.cs
+++ b/SolastaUnfinishedBusiness/Subclasses/PatronSoulBlade.cs
@@ -223,7 +223,7 @@ public sealed class PatronSoulBlade : AbstractSubclass
     // ReSharper disable once UnassignedGetOnlyAutoProperty
     internal override DeityDefinition DeityDefinition { get; }
 
-    private static bool CanWeaponBeEmpowered(RulesetAttackMode mode, RulesetItem item, RulesetCharacter character)
+    public static bool CanWeaponBeEmpowered(RulesetAttackMode mode, RulesetItem item, RulesetCharacter character)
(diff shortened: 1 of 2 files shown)


SolastaUnfinishedBusiness/Subclasses/PatronSoulBlade.cs
```diff
@@ -223,7 +223,7 @@ public PatronSoulBlade()
     // ReSharper disable once UnassignedGetOnlyAutoProperty
     internal override DeityDefinition DeityDefinition { get; }
 
-    private static bool CanWeaponBeEmpowered(RulesetAttackMode mode, RulesetItem item, RulesetCharacter character)
+    public static bool CanWeaponBeEmpowered(RulesetAttackMode mode, RulesetItem item, RulesetCharacter character)
```
ThyWoof, line 226: why Public?

ThyWoof (changes requested):

SolastaUnfinishedBusiness/Subclasses/PatronSoulBlade.cs
BShennette: Err... internal. Sorry, I confused that with protected and thought it wouldn't work.

ThyWoof (approved):

PR #5187, opened 2025-01-31
Way Of Shadow: exclude caster from darkness effects

nelsam:
Unfortunately, I have no Windows PC (everything I own runs linux), so as far as I know I can't compile this and test it. Based on nothing more than gut instinct, I tried to track down how to exclude the Way of Shadow monk from being blinded by their own darkness. I have only the faintest glimmer of hope that it will work.

---

Following the 2024 Warrior of Shadow (which seems to be what the current Way of Shadow subclass is built around), the monk should be able to see through their own Darkness ability.

> You can see within the spell’s area when you cast it with this feature.

This adds '.ExcludeCaster()' to the Way of Shadow's Darkness ability, to match the 2024 rules.


ThyWoof:
unfortunately not that simple.

nelsam:
Cool, thanks for taking the time to look anyway. I made the change before I realized that there's (afaict) no way to build on linux, and I thought I might as well submit it. I hope I didn't waste too much of your time.

ThyWoof:
It’s forcing me to think out of the box again. Good challenge.

